PR #18, opened 2020-10-25
Added a small interactive script that simplifies the process of using…

Floha258:
… multiple options.

Also changed the readme, which might contain some typos now, so please check that again xD


lepelog:
I would still want to display the available command line options, in case people prefer using that. Maybe just put the explanation of the script in a seperate section?

PR #25, opened 2020-10-31
Put Key Pieces and Crystals earlier in logic

azer67:
It looks like Crystals are still broken though. 5 packs don't appear in the spoiler log and I've never seen Batreaux have progress except for 5 and 10 crystals checks.

### DIFF
--- a/SS Rando Logic - Item Location.yaml
+++ b/SS Rando Logic - Item Location.yaml
@@ -194,62 +194,62 @@ Skyloft - Cleaning Crystal:
 
 #  Batreaux Rewards
 Skyloft - Batreaux 5 Crystals:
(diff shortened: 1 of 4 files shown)


lepelog:
It looks like gratitude crystals are already filtered out before, at make_useless_progress_items_nonprogress. Because 5 Gratitude Crystals never shows up in any requirements, since it's treated as 5 single crystals. Crystals lying around aren't relevant here though, since they are forced to be vanilla anyways

azer67:
I fixed crystals via macros. I tried to change "make_useless_progress_items_nonprogress" first, but then there was also a problem when calculating progressin spheres. At least now there can be progress at high Batreaux rewards!

PR #29, opened 2020-11-08
Various changes

azer67:
- Ancient Sea Chart unlocks Ship Dowsing
- Removed talk triggers from Mogmas in Skyview 2 (except the first one)
- Removed Fi Text in front of LMF Boss Door
- Removed a soil patch in the tree eye room of Skyview 2 that can block the box

### DIFF
--- a/SS Rando Logic - Item Location.yaml
+++ b/SS Rando Logic - Item Location.yaml
@@ -184,13 +184,13 @@ Skyloft - Cleaning Crystals:
   Paths:
     - event/123-Town5/186
     - oarc/F016r/l0
-#Skyloft - Owlan Crystals:
-#  Need: Can Access Most of Faron Woods & Bomb Bag
-#  original item: 5 Gratitude Crystals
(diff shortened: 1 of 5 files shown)


azer67:
I don't know very much about github, the goddess crest removal in the Goddess Statue was done by pepper on master, but also by this branch, so make sure it's not getting duplicated or something

azer67:
Owlan's quest now completely works except he doesn't give the item at the end. I'm not sure why so I left it commented in the item location file

azer67:
Alright now Owlan is randomized and works (thx pepper). The conflict is because pepper did half of the fix already but this pull request does the rest.
Also fixed the Academy billboard, sort of. I think there is still a problem when having LMF, FS, AC and ET (because "Ancient Cistern, Earth Temple" is one character too long), but I don't really know how to fix it.

lepelog:
Very nice, I changed the fix for the length a bit, so it breaks them up at whitespaces dynamically

PR #34, opened 2020-11-11
Race Mode puts items not required for required dungeons into non progress pool

azer67:
This patch was done like an elephant walking on butter. So probably don't merge and fix it differently, or merge and modify it greatly. Feel free to use this PR as an exemple (I mean, it looks like this still works lol)

This PR shouldn't have conflicts with the other logic one.


azer67:
I see you did it, so I'll close this

PR #37, opened 2020-11-13
More patches

azer67:
- Start with updated Sand Sea Map (good for casuals to know the sand sea is open)
- Untrigger Fledge on top of academy with bow (to prevent duplicated Fledge)
- Delete Wryna outside skyloft (she's always inside, kind of a duplicate)
- Delete both just sayin' mogmas in front of ET (you can try to talk to them and they have no text so it's weird)
- Delete 3 fi texts in sandship (no more long fi cutscenes after defeating Scervo)

### DIFF
--- a/patches.yaml
+++ b/patches.yaml
@@ -1270,6 +1270,12 @@ D100: # Skyview
     layer: 2
     room: 7
     objtype: OBJ
+  - name: Remove initial Fi text in Skyview 1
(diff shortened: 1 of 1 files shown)


azer67:
Removed the initial fi text in Skyview 1 (it was already done in Skyview 2), and the fi text in the earth temple boss room

PR #88, opened 2021-05-11
Group all difficulties of glitchless, group macros and item requirements

yannl35133:
I managed to put everything together, and it doesn't crash (on the first 100 seeds at least).
I haven't tested the soundness of the seeds.

Also, I based the commit off of gui-redesign, but creating another branch may be better


asm/patch_diffs/custom_funcs_diff.txt
```diff
@@ -38,7 +38,13 @@ main.dol:
...
       0x2E, 0x38, 0x60, 0x00, 0x05, 0x83, 0xC1, 0x00, 0x08, 0x83, 0xE1, 0x00, 0x0C,
       0x80, 0x01, 0x00, 0x14, 0x7C, 0x08, 0x03, 0xA6, 0x38, 0x21, 0x00, 0x10, 0x4E,
-      0x80, 0x00, 0x20]
+      0x80, 0x00, 0x20, 0x81, 0x8D, 0xC3, 0x4C, 0x81, 0x8C, 0x00, 0x04, 0x6D, 0x8C,
+      0xFF, 0xFF, 0x2C, 0x0C, 0x0F, 0xFF, 0x41, 0x82, 0x00, 0x38, 0x81, 0x8D, 0xBB,
+      0xBC, 0x81, 0x8C, 0x00, 0x00, 0x89, 0x6C, 0x53, 0xAD, 0x2C, 0x0B, 0x00, 0x01,
+      0x41, 0x82, 0x00, 0x24, 0x39, 0x8C, 0x53, 0xC0, 0x89, 0x6C, 0x53, 0xCD, 0x2C,
+      0x0B, 0x00, 0x01, 0x41, 0x82, 0x00, 0x14, 0x39, 0x8C, 0x53, 0xC0, 0x89, 0x6C,
+      0x53, 0xCD, 0x2C, 0x0B, 0x00, 0x01, 0x4C, 0x82, 0x00, 0x20, 0x4B, 0xC7, 0x7B,
+      0x68]
```
lepelog, line 47: Please revert the merge of the fix-bit-crashes branch, that merge should be done separately and also will need an ui option as well

asm/patches/ss_necessary.asm
```diff
@@ -343,15 +343,15 @@ nop ; do not need spiral charge to spawn
 .org 0x458
 li r3, 0 ; load a null pointer for the skyloft exit, so the other custom code can take care of it
 
-; this patch branches to the custom trigger_exit_one function, when a scenechange is needed to outside of skykeep
+; this patch branches to the custom trigger_exit_one function, when a scenechange is needed to outside of sky keep
```
lepelog, line 346: If you're already fixing those, make them all uppercase

patches.yaml
```diff
@@ -1858,7 +1858,7 @@ D201: # FS A
     room: 0
     objtype: OBJ
 D201_1: # FS B
-  - name: Clawshot Target for 2nd small Key
```
lepelog, line 1861: I feel like "Clawshot Target" can stay as it is?

ssrando
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
lepelog, line 1: This was probably committed by accident, remove

lepelog (changes requested):
Looks good. I guess `SS Rando Logic - Item Location.yaml` and `SS Rando Logic - Macros.yaml` can be deleted, no need to keep them around since they are replaced with new files.
I've brought this up before and came to the conclusion, that "glitchless logic" and glitched logic should be 2 separate files

asm/patches/ss_necessary.asm
yannl35133: The space is the most important part, and there's no capital letters here, should I really ?

PR #92, opened 2021-05-20
New logic

yannl35133:
Messed up with my repository, I'm not very good with git yet


hints.yaml
```diff
@@ -14,7 +14,7 @@ Sky - Gossip Stone in Bamboo Island:
   textfile: 004-Object
   textindex: 11
   Need: Nothing
-Thunderhead - Gossip Stone on Bug Heaven:
+Thunderhead - Gossip Stone on Bug Rock:
```
lepelog, line 17:
```suggestion
Thunderhead - Gossip Stone near Bug Heaven:
```
It's actually not on the island, also I prefer Bug Heaven over Bug Rock

hints.yaml
```diff
@@ -38,23 +38,23 @@ Eldin Volcano - Gossip Stone in Thrill Digger Cave:
...
+  Need: Can Get Past First Frog
+Volcano Summit - Gossip Stone in Waterfall Area:
   textfile: 004-Object
   textindex: 26
   Need: Can Access Volcano Summit & Clawshots
 Lanayru Desert - Gossip Stone in Temple of Time Area:
   textfile: 004-Object
   textindex: 25
-  Need: Can Access Second Part of Lanayru & Hook Beetle
+  Need: Can Access Front of Temple of Time & Hook Beetle
```
lepelog, line 52: This does not need Hook Beetle with the precise slingshot shot and also for glitched logic you can get there with beating the dungeon inside LMF

logic/hints.py
```diff
@@ -10,30 +10,30 @@
...
+    "Sky - Kina's Crystals",
     "Skyloft - Peater/Peatrice's Crystals",
     'Batreaux - 80 Crystals',
 ]
 
 SOMETIMES_LOCATIONS = [
-    'Lanayru Sand Sea - Roller Coaster Minigame',
-    'Skyloft Academy - Pumpkin Archery - 600 Points',
+    'Lanayru Sand Sea - Rickety Coaster',
+    'Knight Academy - Pumpkin Pull - 600 Points',
```
lepelog, line 20: I think this should stay as Pumpkin Archery

logic/hints.py
```diff
@@ -10,30 +10,30 @@
...
     'Sky - Lumpy Pumpkin Harp Minigame',
-    'Sky - Fun Fun Island Minigame',
-    'Thunderhead - Bug Island minigame',
+    "Sky - Dodoh's High Dive - 500 Rupees",
```
lepelog, line 22: I think Fun Fun Island Minigame can stay as it is

logic/hints.py
```diff
@@ -10,30 +10,30 @@
...
+    'Thunderhead - Bug Heaven - 10 Bugs in 3 Minutes',
```
lepelog, line 23: Change that back as well

randomizer-changes.md
```diff
@@ -71,7 +71,7 @@
 ### Fire Sanctuary
 - The first cutscene with Guld has been removed
 - The initial cutscene of the first trapped mogma was removed to prevent a possible softlock when activating it out of order
-- A clawshot target was added near the first trapped mogma chest. This is so you can escape from the corner without dying without gust bellows
+- A clawshots target was added near the first trapped mogma chest. This is so you can escape from the corner without dying without gust bellows
```
lepelog, line 74: clawshot target

lepelog (changes requested):
Looks good, just a few things

Also split the logic into glitchless and glitched logic, I'll review the logic files after that. Since the tricks for glitched and glitched should be entirely separate imo (and glitched logic should behave as having all glitchless tricks enabled), they should also be separate on the UI, although that would be a job for @cjs8487

hints.yaml
yannl35133: Those were unused

yannl35133:
What would be the benefits of a split between Glitched and Glitchless, apart from code duplication ?

What should the basis of check names be ? I considered official names as a pretty solid one.

I also consider that including score requirements in minigame checks (that allow it) is a good idea. I forgot to add the requirement of Rickety Coaster, I'll add it.

Should I solve the conflicts with the current gui-redesign ?

lepelog:
For check names, it just really matters that it's recognizable. Like, if you have to think "is this the right check", then the name is probably bad. Imo it's not in the scope of this PR to fix check names, I already gave feedback to some I want to be changed back and the rest I don't really mind. Fixing typos in check names is fine. I don't really know how to properly coordinate check name changes (since there are a lot of names we always wanted to change), probably in discord somehow. After this is merged, you are also welcome to submit a PR with name changes.

Not sure if having the required points/time in the check name is a good idea, but as I said above, it's out of scope and should be discussed in discord. Generally I prefer shorter check names, cause they fit on in game textboxes more easily, but cj is working on removing that restriction so it's something that we can think about.

Yes, please resolve the conflict

yannl35133:
Conflict resolved, name changes reverted (I'll make another PR)

lepelog:
The only thing holding me back from merging this is that glitched and glitchless tricks are not separated. I refuse to merge this when glitchless tricks can be disabled with BiT active, we already agreed on that glitched logic should have all glitched tricks implicitly enabled. I'm fine with merging the one file solution now until I have time to refactor it.

yannl35133:
I don't entirely agree, I don't see why it is necessary, but I won't oppose it (as fiercely as the one file requirements)

PR #95, opened 2021-05-26
Water Dragon Healing Patch & RBM fixes

Battlecats59:
Added an option to turn on healing the water dragon, which requires sacred water to access Ancient Cistern
Fixed various layers and objects to allow for more instances of RBM

### DIFF
--- a/options.yaml
+++ b/options.yaml
@@ -256,6 +256,14 @@
   help: Controls the conditions for opening LMF. Nodes requires activating the 3 nodes as in vanilla,
     Hook Beetle will raise LMF once the Hook Beetle is obtained.
   ui: option_open_lmf
+- name: Heal Water Dragon
+  command: heal-water-dragon
+  type: boolean
+  bits: 2
+  default: false
+  help: If checked, getting sacred water from Skyview Spring is required to heal the Water Dragon to
+    enter Ancient Cistern.
+  ui: option_heal_water_dragon
 - name: Horde
   command: horde
   type: boolean
--- a/patches.yaml
+++ b/patches.yaml
@@ -28,6 +28,8 @@ global:
     - 519 # Updated Sand Sea map
     - 115 # Full Lake Floria map
     - 180 # Thrill digger entrance CS
+    - 148 # Link looking around cutscene in Sealed Temple (for bokos in BtT)
+    - 64 # Water Dragon quest started
   startitems: # careful, items are NOT progressive here (only set item flag, no text), storyflag for upgrade has to be set manually
     - 15 # sailcloth
   startsceneflags: # currently only for skyloft
@@ -102,6 +104,12 @@ B100:
     room: 0
     objtype: STAG
 B100_1: # skyview spring
+  - name: Layer override
+    type: layeroverride
+    override:
+      - story_flag: -1
+        night: 0
+        layer: 4
   - name: Fi text after skyview CS
     type: objpatch
     id: 0xFC73
@@ -766,12 +774,6 @@ F008r: # Goddess Statue
     layer: 0
     room: 0
     objtype: OBJ
-  - name: Delete Crest
-    type: objdelete
-    id: 0xFC02
-    layer: 0
-    room: 0
-    objtype: OBJ
 F010r: # Isle of songs tower
   - name: Delete Faron Trial Dowsing
     type: objdelete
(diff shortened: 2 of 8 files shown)


options.yaml
```diff
@@ -256,6 +256,14 @@
   help: Controls the conditions for opening LMF. Nodes requires activating the 3 nodes as in vanilla,
     Hook Beetle will raise LMF once the Hook Beetle is obtained.
   ui: option_open_lmf
+- name: Heal Water Dragon
+  command: heal-water-dragon
+  type: boolean
+  bits: 2
+  default: false
+  help: If checked, getting sacred water from Skyview Spring is required to heal the Water Dragon to
+    enter Ancient Cistern.
```
lepelog, line 265: This feels kinda backwards, if I check "Heal Water Dragon" I would expect the dragon to be healed. Maybe "Open Ancient Cistern" would be a better wording?

patches.yaml
```diff
@@ -102,6 +104,12 @@ B100:
     room: 0
     objtype: STAG
 B100_1: # skyview spring
+  - name: Layer override
+    type: layeroverride
+    override:
+      - story_flag: -1
+        night: 0
+        layer: 4
```
lepelog, line 112: Doesn't this skip the scale requirement for getting sacred water? Or was that an intentional decision to make it possible to avoid backtracking?

patches.yaml
```diff
@@ -1109,9 +1107,28 @@ F102_1: # Outside AC
...
+        layer: 3
+      - story_flag: 155 # AC intro cs
+        night: 0
+        layer: 2
+      - story_flag: 36 # Tunic
+        night: 0
+        layer: 1
+      - story_flag: -1
+        night: 0
+        layer: 1
```
lepelog, line 1131: Did you mean to do layer 0 here? A layer 0 entry all at the bottom doesn't do anything btw, cause layer 0 is the default anyways

patches.yaml
```diff
@@ -1632,6 +1664,22 @@ F200: # Eldin main area
     type: oarcadd
     oarc: GossipStone
     destlayer: 0
+  - name: extra Bomb flower # Deathwarp if getting cube via Eldin OoB
```
lepelog, line 1667: I feel like it's intentional that you softlock without bombs if you get there via eldin OoB, not sure if we want to change that since it doesn't come up in normal gameplay without glitches

patches.yaml
```diff
@@ -1120,6 +1137,21 @@ F102_1: # Outside AC
     objtype: OBJ
     object:
       trigstoryfid: 900 # new AC beat flag
+F102_2: # Water Dragon's Lair
+  - name: Layer override
```
lepelog, line 1141: Imo healing the water dragon even if the option is not selected should be possible, so just layer 1 always should be fine

patches.yaml
```diff
@@ -1109,9 +1107,28 @@ F102_1: # Outside AC
...
+    onlyif: Option "heal-water-dragon" Enabled # Water Dragon Quest
+    type: layeroverride
+    override: # change layer 3 trigger, no layer 1
+      - story_flag: 900 # new beat AC
+        night: 0
+        layer: 3
+      - story_flag: 155 # AC intro cs
+        night: 0
+        layer: 2
```
lepelog, line 1125: This is for the inside, so it breaks with entrance rando

lepelog (changes requested):
good stuff, haven't been able to test the changes, but already have a few things to address

patches.yaml
Battlecats59: Yes, it was intentional to avoid backtracking. Would it be better to require scale?

patches.yaml
Battlecats59: For no logic and glitched logic, a bomb flower here would let the player deathwarp after getting the cube with Eldin OoB

patches.yaml
Battlecats59: It's layer 1 to allow RBMing the AC entering cutscene. I just removed the tunic layer because it also set layer 1, so there was no use for it

patches.yaml
lepelog: no, I think this change is fine, just put it as a note in `randomizer-changes.md` so we have it documented

Battlecats59:
Bug notes:
- 2 Skyview Boss Keys in the item pool when boss keys are "Anywhere" or "Vanilla"
- Water dragon disappears (after beating FS?)

Do not merge until these are fixed; currently working on fixing these

PR #103, opened 2021-06-25
Trial rando final touches

Battlecats59:


logic/hints.py
```diff
@@ -84,10 +82,8 @@ def to_gossip_stone_text(self) -> str:
 
     def to_spoiler_log_text(self) -> str:
         zone, specific_loc = Logic.split_location_name_by_zone(self.location_name)
-        trials = [("Skyloft Silent Realm", "Skyloft Trial Gate"), ("Faron Silent Realm", "Faron Trial Gate"), ("Lanayru Silent Realm", "Lanayru Trial Gate"), ("Eldin Silent Realm", "Eldin Trial Gate")]
```
lepelog, line 87: I think it would make more sense to have a `TrialGateItemGossipStoneHint` subclass that deals with item hints for trials to avoid adding that logic to the existing `ItemGossipStoneHint` class where it doesn't belong imo

lepelog (changes requested):
looks really nice, good job. I would just change how the hints are generated then it looks good. But change the PR to point to gui-redesign, the trial-rando branch was just so that I could push the asm stuff

lepelog (approved):

PR #114, opened 2021-09-07
Sword Dungeon Reward

YourAverageLink:
Added an option for forcing a sword to appear at the end of a required dungeon, if possible.

### DIFF
--- a/logic/hints.py
+++ b/logic/hints.py
@@ -5,7 +5,7 @@ from collections import OrderedDict, defaultdict
 from dataclasses import dataclass
 from typing import List
 
-from .constants import POTENTIALLY_REQUIRED_DUNGEONS, SILENT_REALMS, SILENT_REALM_CHECKS
+from .constants import POTENTIALLY_REQUIRED_DUNGEONS, SILENT_REALMS, SILENT_REALM_CHECKS, END_OF_DUNGEON_CHECKS
 from util import textbox_utils
 
 ALWAYS_REQUIRED_LOCATIONS = [
@@ -295,6 +295,8 @@ class Hints:
         hintable_items = HINTABLE_ITEMS.copy()
         item_hints = []
         self.logic.rando.rng.shuffle(hintable_items)
+        for i in self.logic.sworded_dungeon_locations:
+            hinted_locations.append(i)
         for i in range(self.logic.rando.options["item-hints"]):
             hinted_item = hintable_items.pop()
             for location, item in self.logic.done_item_locations.items():
--- a/logic/hints.py
+++ b/logic/hints.py
@@ -5,7 +5,7 @@ from collections import OrderedDict, defaultdict
 from dataclasses import dataclass
 from typing import List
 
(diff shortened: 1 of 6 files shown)


logic/hints.py
```diff
@@ -295,6 +295,8 @@ def do_normal_hints(self):
         hintable_items = HINTABLE_ITEMS.copy()
         item_hints = []
         self.logic.rando.rng.shuffle(hintable_items)
+        for i in self.logic.sworded_dungeon_locations:
+            hinted_locations.append(i)
```
lepelog, line 299:
```suggestion
        hinted_locations.extend(self.logic.sworded_dungeon_locations)
```

logic/logic.py
```diff
@@ -189,6 +190,28 @@ def __init__(self, rando):
...
+            "True Master Sword": 6,
+        }
+        #yes this is defined twice, idk why but it only works when i do that. sorry
+        swords_left = 6 - starting_sword_count[self.rando.options["starting-sword"]]
+        self.sworded_dungeon_locations = []
+        if (self.rando.options["sword-dungeon-reward"] and ("dungeon" not in self.rando.banned_types)):
+            if swords_left < self.rando.options["required-dungeon-count"]:
+                sworded_dungeons = self.rando.rng.sample(self.required_dungeons,k=swords_left)
+            else:
+                sworded_dungeons = self.required_dungeons.copy()
```
lepelog, line 209: no need to copy here

logic/logic.py
```diff
@@ -189,6 +190,28 @@ def __init__(self, rando):
...
+        starting_sword_count = {
+            "Swordless": 0,
+            "Practice Sword": 1,
+            "Goddess Sword": 2,
+            "Goddess Longsword": 3,
+            "Goddess White Sword": 4,
+            "Master Sword": 5,
+            "True Master Sword": 6,
+        }
+        #yes this is defined twice, idk why but it only works when i do that. sorry
```
lepelog, line 202: not sure what this comment is supposed to mean, either clarify or remove it

logic/hints.py
```diff
@@ -5,7 +5,7 @@
 from dataclasses import dataclass
 from typing import List
 
-from .constants import POTENTIALLY_REQUIRED_DUNGEONS, SILENT_REALMS, SILENT_REALM_CHECKS
+from .constants import POTENTIALLY_REQUIRED_DUNGEONS, SILENT_REALMS, SILENT_REALM_CHECKS, END_OF_DUNGEON_CHECKS
```
lepelog, line 8: Doesn't seem like you're actually using the import, so remove it

lepelog (changes requested):
There is an issue with swordless and all dungeons required, then you can't reach any end of the dungeon so the seed softlocks. And with only 1 sword you can only beat LMF, so at least Goddess Sword should be reachable without the end of dungeon checks

Also please run the formatter with `black .`

lepelog (approved):

PR #116, opened 2021-09-11
Use argparse for parsing command line arguments

lepelog:
Makes for a nicer help, it also has some checking functionality already built in

### DIFF
--- a/randoscript.py
+++ b/randoscript.py
@@ -13,151 +13,110 @@ def get_ranges(start, end, parts):
         yield (int(start + step * i), int(start + step * (i + 1)))
 
 
-if "NOGIT" in VERSION:
-    print("WARNING: Running from source, but without git, this is highly discouraged")
+def main():
+    if "NOGIT" in VERSION:
+        print(
+            "WARNING: Running from source, but without git, this is highly discouraged"
+        )
 
-# add options
-parser = argparse.ArgumentParser(
-    description="Skyward Sword Randomizer Version " + VERSION
-)
-parser.add_argument(
-    "--permalink",
-    help="Specify a permlink, which includes the settings. This is set first, other options may override these settings",
-)
-parser.add_argument(
-    "--placement-file",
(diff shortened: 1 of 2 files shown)


randoscript.py
```diff
@@ -79,17 +16,96 @@ def get_ranges(start, end, parts):
...
-    else:
-        cmd_line_args[option_name[2:]] = arg_parts[1]
-plcmt_file_name = cmd_line_args.pop("placement-file", None)
+# add options
+parser = argparse.ArgumentParser(
+    description="Skyward Sword Randomizer Version " + VERSION
+)
+parser.add_argument(
+    "--permalink",
+    help="Specify a permlink, which includes the settings. This is set first, other options may override these settings",
```
yannl35133, line 25: perm*a*link

randoscript.py
```diff
@@ -79,17 +16,96 @@ def get_ranges(start, end, parts):
...
+    type=int,
+    dest="bulk_threads",
+)
+
+parsed_args = parser.parse_args()
+if parsed_args.version:
+    print(VERSION)
+    exit(0)
+options = Options()
+if not parsed_args.permalink is None:
```
yannl35133, line 90: `parsed_args.permalink is not None` is the Python way (if I understood correctly your code)

randoscript.py
```diff
@@ -79,17 +16,96 @@ def get_ranges(start, end, parts):
...
+for optname, opt in OPTIONS.items():
+    optval = parsed_args.__getattribute__(optname.replace("-", "_"))
+    if not optval is None:
+        value, validation_errors = Options.parse_and_validate_option(optval, opt)
+        if len(validation_errors) > 0:
+            all_errors.extend(validation_errors)
+        else:
+            options.set_option(optname, value)
+
+if len(all_errors) > 0:
```
yannl35133, line 102: `if all_errors:` already checks if it's non-empty

yannl35133 (commented):
I still think that wrapping everything in a main function would be better, but I may be wrong

PR #121, opened 2021-09-25
show if SK is required on the academy billboard

lepelog:


lepelog:
Most people are against this change

PR #129, opened 2021-10-06
Race integrity patches and filename parsing support

Battlecats59:


gamepatches.py
```diff
@@ -1438,6 +1443,35 @@ def add_stone_hint_patches(self):
...
 
+    def add_race_integrity_patches(self):
+        self.eventpatches["599-Demo"].append(
+            {
+                "name": "Race Integrity Patch for Fi",
+                "type": "textpatch",
+                "index": 153,
+                "text": make_mutliple_textboxes(
+                    [
+                        "Congratulations, Master <heroname>.\nSeed: {0}\nHash: {1}".format(
```
lepelog, line 1454:
```suggestion
                        f"Congratulations, Master <heroname>.\nSeed: {self.rando.seed}\nHash: {self.placement_file.hash_str}"```
I would prefer f format

File: gamepatches.py [hunk 2]
```diff
@@ -1438,6 +1443,35 @@ def add_stone_hint_patches(self):
...
+                            self.rando.seed, self.placement_file.hash_str
```
lepelog, line 1455: Maybe remove the seed? This will error when using a plandomized placement_file

PR #133, opened 2021-10-23
Nodes as separate regions from Lanaryu Desert

YourAverageLink:


cjs8487:
Closing as a wontfix since this was determined to be a nonissue and interferes with ER logic

PR #142, opened 2021-11-16
Cannot install wit clearer error message

yannl35133:


cjs8487 (approved):
lgtm, assuming there were no other issues with this that prevented us from merging it way sooner

PR #144, opened 2021-11-18
use certifi in build

lepelog:
build: https://nightly.link/lepelog/ssrando/actions/runs/1478253756

### DIFF
--- a/ssrando.spec
+++ b/ssrando.spec
@@ -49,6 +49,14 @@ else:
 with open('version-with-git.txt','w') as f:
   f.write(VERSION)
 
+hiddenimports = []
+
+try:
+  import certifi
+  hiddenimports.append('certifi')
+except ModuleNotFoundError:
+  pass
+
 a = Analysis(['randoscript.py'],
              pathex=[],
              binaries=[],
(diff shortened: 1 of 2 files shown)


ssrando.spec
```diff
@@ -49,6 +49,14 @@ else:
 with open('version-with-git.txt','w') as f:
   f.write(VERSION)
 
+hiddenimports = []
+
+try:
+  import certifi
+  hiddenimports.append('certifi')
+except:
```
yannl35133, line 57: Empty excepts look bad, why not have
```suggestion
except ModuleNotFondError:
```

lepelog: It doesn't do any error handling anyways, so I don't see the benefit of adding which error gets thrown here. Plus I don't have to worry about if there is any other error that can possibly be thrown

yannl35133: Catching all errors (if you don't re-raise) is usually bad practice; you don't want to assume all errors should be ignored.
In this case, imagine for some reason importing certifi blocks (eg. infinite loop)
Then, if the user interrupts, the rando resumes in the `except` case, while the user just wanted to completely interrupt everything

lepelog:
didn't seem to help

PR #147, opened 2021-12-19
Add option to shuffle required dungeon entrances

YourAverageLink:


lepelog (changes requested):
Looks good :+1:

logic/logic.py
```diff
@@ -321,42 +334,38 @@ def randomize_entrance_connections(self):
...
-                ]
-            )
-            assert len(dungeons) == 0, "Not all dungeons linked to an entrance"
-            # TODO: do checks here if in the future some connections aren't possible
-            connections_invalid = False
-        return entrance_connections
+                    unreq_indices.append(index)
+
+            if not self.rando.options["skip-skykeep"]:
+                req_indices.append(dungeons.index("Sky Keep"))
```
lepelog, line 356: SK index needs to be added to the unreq indices if it's not required

lepelog (changes requested):

logic/logic.py
yannl35133: I thought the desired behaviour was that if SK was skipped, it was left in place and made entirely dead

YourAverageLink: Well, now that we decided unrequired dungeons should be shuffled, sky keep should be shuffled among the unrequired dungeons if it is skipped

PR #150, opened 2022-01-22
open et

lepelog:
just delete the door actor to avoid messing with flags (since it's the shed flag)


lepelog:
someone still needs to put it on the ui, I'm bad at that

cjs8487:
I can do the UI once this is merged

PR #153, opened 2022-02-04
Cutscene skips

YourAverageLink:
Shortens/removes various other cutscenes in dungeons or side quests


lepelog:
Is there a reason for this still being draft? You can always add more cs skips in future PRs

YourAverageLink:
I only had it as draft from your suggestion; I guess enough time has passed and testing been done to make this ready for review.

PR #156, opened 2022-02-21
Custom music support

Battlecats59:
Patches music file length to 2GB for every file to prevent custom tracks from cutting off


cjs8487 (dismissed):
3/4 success rate is not sufficient to prove that this works without fail, especially when failures completely corrupt extracts

Battlecats59:
CJ try to generate a seed with this version using the "Allow Custom Music" option, without custom music. That option should, in no way, corrupt your extract, it must be something else, related the music, that most likely caused it - then if that works you can try with the music and see if that corrupts your extract.

Battlecats59:
Update: CJ's issue was determined to be not related to this patch, so this should be good to go

musicrando.py
```diff
@@ -84,7 +84,10 @@ def music_rando(self):
             ):
                 track_len = self.musiclist[original_track]["audiolen"]
             else:
-                track_len = self.musiclist[new_track]["audiolen"]
+                if self.placement_file.options["allow-custom-music"]:
+                    track_len = 2147483647  # 0x7F FF FF FF - 2GB
```
lepelog, line 88:
```suggestion
                    track_len = 0x7FFFFFFF # - 2GB
```
Makes more sense as hex

lepelog (changes requested):

PR #157, opened 2022-02-27
hacky no logic entrance rando

lepelog:
Probably has a lot of problems, the yaml file for the entrance definitions is also definitely not final.

WARNING: after this, you have to copy the entire `Stages` folder from `actual-extract` to `modified-extract` since this also touches stages the normal rando never touches.


harmjan387:
https://nightly.link/lepelog/ssrando/workflows/build.yaml/no-logic-er/dist%20windows-latest

PR #158, opened 2022-03-09
make ToT always layer 2, locate machi after SV, unlock endurance potion with raising LMF

lepelog:
Preparation for ER to get rid of the concept of "dungeon inside dungeon" (which doesn't make sense with full ER)


lepelog:
Link for test builds: https://nightly.link/lepelog/ssrando/workflows/build.yaml/no-dungeon-in-dungeon/dist%20windows-latest

PR #162, opened 2022-03-28
Better Hints

cjs8487:
This will need some extensive testing, and the distros could probably be refined a bit further, but this should be it _finally_. For info on writing/reading distro files check the readme in the distribution directory. Note that this is **not** peramlink compatible with previous versions


hints/hint_distribution.py
```diff
@@ -0,0 +1,434 @@
...
+        self.barren_overworld_zones = []
+        self.placed_ow_barren = 0
+        self.barren_dungeons = []
+        self.placed_dungeon_barren = 0
+        self.prev_barren_type = None
+        self.hinted_locations = []
+        self.weighted_types = []
+        self.weights = []
+        self.hintable_items = []
+        self.ready = False
```
lepelog, line 106: unused?

hints/hint_distribution.py
```diff
@@ -0,0 +1,434 @@
...
+                    self.hints.extend(
+                        [self._create_sometimes_hint()] * curr_type["copies"]
+                    )
+            elif type == "sots":
+                for i in range(curr_type["fixed"]):
+                    try:
+                        self.hints.extend(
+                            [self._create_sots_hint()] * curr_type["copies"]
+                        )
+                    except:
```
lepelog, line 281: except without an explicit exception is discouraged I think, because this would also catch the user trying to stop the rando program with `Ctrl-C`

hints/hint_distribution.py
```diff
@@ -0,0 +1,434 @@
...
+            if item in self.hintable_items:
+                self.hintable_items.remove(item)
+        for item in self.logic.starting_items:
+            if item in self.hintable_items:
+                self.hintable_items.remove(item)
+        self.logic.rando.rng.shuffle(self.hintable_items)
+
+        needed_fixed = []
+        for type in self.distribution.keys():
+            if self.distribution[type]["fixed"] > 0:
```
lepelog, line 260: Wouldn't it also be reasonable to set fixed to 0 to disable a specific hint type? I think I would omit it from the distribution json when it doesn't use fixed and use `if "fixed" in self.distribution[type]:`here

hints/hint_distribution.py
lepelog: I think I would prefer if `self._create_sots_hint` would return `None` if the requested hint couldn't be generated

Then we can do
```python
if hint := self._create_sots_hint():
  self.hints.extend([hint] * curr_type["copies"])
```

lepelog (changes requested):
Great job with this, haven't looked over everything yet so might comment again

hints/hint_distribution.py
cjs8487: The design of distribution files are to be at least somewhat readable to someone who wants to know what the details of the distribution without knowing all the possible fields. Explicitly declaring that there are 0 fixed hints means that you can tell the details of the distribution at a glance without having to think about the fact that it was omitted

hints/distributions/README.md
```diff
@@ -0,0 +1,40 @@
+# SSR Hint Distributions
+
+Hint distributions define everything about gossip stone hints are placed by the randomizer. Distributions have the ability to fine tune exactly how hints are placed in the world, by banning specific stones, adding and removing items and locations from their respecitve pools, and more.
```
YourAverageLink, line 3: I suggest adding "how":
"Hint distributions define everything about how gossip stone hints are placed by the randomizer."

hints/distributions/README.md
```diff
@@ -0,0 +1,40 @@
...
+  - Limits the number of dungeon SotS hints that can be placed. They will attempt be replaced by a different SotS hint if a hint past the limit would be placed.
+- dungeon_barren_limit (number)
+  - Limits the number of barren dungeon hints that can be placed. They will attempt to be replaced by an overworld barren hint if another hint past the limit would be placed.
+- distribution (name: distribution_data (see below))
+  - Defines the specifics of the distribution
+
+### Distribution Data
+- order: defines what order hints should have their fixed quantities generated. Types with the same order will be resolved in the order they are listed in the distribution
+- weight: The relative chance that this hint type is randomly selected to be placed
+- fixed: Generates the specified number of hints prior to randomly selecting additional hints by weight. Note that this **does not** guarantee that all these hints will be placed dpeending on the rest of the distribution
```
YourAverageLink, line 39: dpeending cjTypo

hints/hint_distribution.py
```diff
@@ -0,0 +1,461 @@
...
+            if type["copies"]:
+                self.hints.extend([hint] * (type["copies"] - 1))
+            return hint
+        elif next_type == "random":
+            hint = self._create_random_hint()
+            if type["copies"]:
+                self.hints.extend([hint] * (type["copies"] - 1))
+            return hint
+        # junk hint is the last possible type and also a fallback
+        return EmptyGossipStoneHint(None, None, False, self.junk_hints.pop())
```
lepelog, line 355:
```suggestion
        elif next_type == "random":
            hint = self._create_random_hint()
        else:
            # junk hint is the last possible type and also a fallback
            return EmptyGossipStoneHint(None, None, False, self.junk_hints.pop())
        if type["copies"]:
            self.hints.extend([hint] * (type["copies"] - 1))
        return hint
```
deduplicate the copy mechanism

hints/hint_distribution.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def next_hint(self) -> GossipStoneHint:
+        if len(self.hints) > 0:
+            return self.hints.pop()
+        [next_type] = self.rng.choices(self.weighted_types, self.weights)
+        type = self.distribution[next_type]
+        if next_type == "sometimes":
+            hint = self._create_sometimes_hint()
+            if type["copies"]:
+                self.hints.extend([hint] * (type["copies"] - 1))
```
lepelog, line 332: If I understand correctly, this could end up not generating a copy if this is the last hint that gets placed?

hints/hint_distribution.py
```diff
@@ -0,0 +1,461 @@
...
+        elif next_type == "random":
+            hint = self._create_random_hint()
+            if type["copies"]:
+                self.hints.extend([hint] * (type["copies"] - 1))
+            return hint
+        # junk hint is the last possible type and also a fallback
+        return EmptyGossipStoneHint(None, None, False, self.junk_hints.pop())
+
+    def _create_sometimes_hint(self):
+        hint = self.sometimes_hints.pop()
```
lepelog, line 358:
```suggestion
        if not self.sometimes_hints:
            return None
        hint = self.sometimes_hints.pop()
```

hints/hint_distribution.py
```diff
@@ -0,0 +1,461 @@
...
+        while hint in self.hinted_locations:
+            hint = self.sometimes_hints.pop()
```
lepelog, line 360:
```suggestion
            if not self.sometimes_hints:
                return None
            hint = self.sometimes_hints.pop()
```

hints/hint_distribution.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def _create_sots_hint(self):
+        if len(self.sots_locations) == 0:
+            return None
+        zone, loc, item = self.sots_locations.pop()
+        while loc in self.hinted_locations:
+            if len(self.sots_locations) == 0:
+                return None
+            zone, loc, item = self.sots_locations.pop()
+        self.hinted_locations.append(loc)
```
lepelog, line 377: shouldn't this be directly before the return?

hints/hint_types.py
```diff
@@ -0,0 +1,125 @@
...
+    def __hash__(self):
+        return hash(self.location + self.item)
+
+
+@dataclass
+class LocationGossipStoneHint(GossipStoneHint):
+    location_string: str
+
+    def to_gossip_stone_text(self) -> List[str]:
+        return [f"They say that {self.location_string} <y<{self.item}>>"]
```
lepelog, line 60:
```suggestion
        return [f"They say that <r<{self.location_string}>> <y<{self.item}>>"]
```
shouldn't the location part be colored like it used to be?

YourAverageLink: This colors everything in the location string, including articles and the final verb (like reveals) which CJ thinks should not be highlighted (and I agree).

lepelog: hm that's unfortunate, the coloring makes it a lot easier to read imo but I agree as well that only the location name should be highlighted. Maybe we can revisit it in the future

YourAverageLink: I'm currently adding coloring to every hint name

hints/hint_distribution.py
cjs8487: Yes, this applies to all hint types. Not really sure of a good solution unfortunately. Open to suggestions if you have any

hints/hint_distribution.py
```diff
@@ -0,0 +1,450 @@
...
+
+        for loc in self.added_locations:
+            location = loc["location"]
+            if loc["type"] == "always":
+                if location in always_hints:
+                    continue
+                always_hints.append(location)
+                if location in sometimes_hints:
+                    if location in sometimes_hints:
+                        continue
```
lepelog, line 151: this is always true, I guess this should be removed?

hints/hint_distribution.py
```diff
@@ -0,0 +1,450 @@
...
+                    sometimes_hints.remove(location)
+            elif loc["type"] == "sometimes":
+                sometimes_hints.append(location)
+                if location in sometimes_hints:
+                    always_hints.remove(location)
```
lepelog, line 156:
```suggestion
                if location not in sometimes_hints:
                    sometimes_hints.append(location)
```
not sure if that's what you want to do here

lepelog (changes requested):
:+1: for updating the changelog

lepelog (approved):
lgtm :+1:

PR #164, opened 2022-04-02
Text patches / more CS trimming

YourAverageLink:


patches.yaml
```diff
@@ -1241,13 +1241,17 @@ F102_1: # Outside AC
...
     room: 0
     objtype: OBJ
-    object:
-      trigstoryfid: 900 # new AC beat flag
+  - name: Water Dragon after AC
+    type: objdelete
+    id: 0xFC5A
+    layer: 3
+    room: 0
+    objtype: OBJ
```
lepelog, line 1254: this still tries to play the cutscene after cistern, I fixed this on another branch:
```yaml
  - name: Remove CS after cistern
    type: objpatch
    index: 6
    room: 0
    objtype: PLY
    object:
      play_cutscene: -1
```

lepelog (approved):

PR #172, opened 2022-04-15
write hints as json in json spoiler log

lepelog:


YourAverageLink (commented):
Maybe add a distinction between always, sometimes, and random hints as opposed to just all location, if possible

cjs8487 (approved):

PR #174, opened 2022-04-23
spawn training fledge without wryna

lepelog:


lepelog:
https://nightly.link/lepelog/ssrando/workflows/build.yaml/fledge-wryna

PR #180, opened 2022-04-30
Rework cube hints & add item location/zone switch

YourAverageLink:
Change the interaction such that barren hints factor their goddess cubes in as well and sots hints pointing to goddess chests display the zone of the associated cube instead of the chest. Also adds an option for modifying the sots hint to say it involves a cube.

### DIFF
--- a/hints/hint_distribution.py
+++ b/hints/hint_distribution.py
@@ -447,7 +447,12 @@ class HintDistribution:
             return None
         location, item = self.rng.choice(locs)
         self.hinted_locations.append(location)
-        return ItemGossipStoneHint(location, item, True)
+        if self.logic.rando.options["precise-item"]:
+            return ItemGossipStoneHint(location, item, True, None)
+        zone_override, _ = self.logic.split_location_name_by_zone(location)
+        if "Goddess Chest" in location:
+            zone_override = self.logic.rando.item_locations[location]["cube_region"]
+        return ItemGossipStoneHint(location, item, True, zone_override)
 
     def _create_random_hint(self):
         all_locations_without_hint = self.logic.filter_locations_for_progression(
--- a/hints/hint_distribution.py
+++ b/hints/hint_distribution.py
@@ -8,7 +8,6 @@ from logic.constants import (
     POTENTIALLY_REQUIRED_DUNGEONS,
     ALL_DUNGEON_AREAS,
     SILENT_REALM_CHECKS,
-    GODDESS_PRECISE_ZONES,
 )
 from logic.logic import Logic
 from paths import RANDO_ROOT_PATH
@@ -107,7 +106,6 @@ class HintDistribution:
         self.junk_hints = []
         self.sometimes_hints = []
         self.hint_descriptors = []
-        self.goddess_cube_zones = GODDESS_PRECISE_ZONES
 
     def read_from_file(self, f):
         self._read_from_json(json.load(f))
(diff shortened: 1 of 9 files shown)


options.yaml
```diff
@@ -544,4 +544,10 @@
   default: Balanced
   help: Sets the distribution of hints. Distributions describe how many hints are placed and in what order, as well
     as specifics about what additional items and locations can and cannot be hinted
-  ui: option_hint_distribution
\ No newline at end of file
+  ui: option_hint_distribution
```
cjs8487, line 547: put the option before the distribution option, the distro option is intended to always be the last option for maximum backwards compatibility of permas when the list changes for whatever reason

cjs8487 (changes requested):

options.yaml
YourAverageLink: It gets put before the distro in the permalink anyway it seems

cjs8487: It shouldn't, if it does that is very strange. It may just look like it is depending on where it is in the bits for those characters

hints/hint_distribution.py
```diff
@@ -368,6 +372,17 @@ def _create_sots_hint(self):
         elif zone in ALL_DUNGEON_AREAS:
             self.sots_dungeon_placed += 1
         self.hinted_locations.append(loc)
+        if "Goddess Chest" in loc:
+            zone = self.goddess_cube_zones[loc]
+            # place cube sots hint & catch specific zones and fit them into their general zone (as seen in the cube progress options)
+            if self.logic.rando.options["cube-sots"]:
+                if zone == "Skyview":
+                    zone = "Faron Woods"
+                if zone == "Mogma Turf":
```
lepelog, line 381:
```suggestion
                elif zone == "Mogma Turf":
```

hints/hint_distribution.py
```diff
@@ -106,6 +107,7 @@ def __init__(self):
         self.junk_hints = []
         self.sometimes_hints = []
         self.hint_descriptors = []
+        self.goddess_cube_zones = GODDESS_PRECISE_ZONES
```
lepelog, line 110: this member is not needed at all, just use `GODDESS_PRECISE_ZONES` directly

logic/constants.py
```diff
@@ -131,6 +131,35 @@
     "True Master Sword": 6,
 }
 
+GODDESS_PRECISE_ZONES = {
+    "Central Skyloft - Bazaar Goddess Chest": "Lanayru Sand Sea",
```
lepelog, line 135: I think it would be better to have the information, in what region the cube is, added to `checks.yaml`

lepelog (changes requested):
a few nitpicks

logic/constants.py
YourAverageLink: Fair enough, I was just worried that it would mean having to iterate through all of checks.yaml each time i need to fetch the cube zone

lepelog: You could do `self.rando.item_locations[loc]["cube_region"]`

YourAverageLink:
This is somewhat off-topic for the rest of this PR but it is just an option being added so it shouldn't conflict with anything else.

hints/hint_types.py
```diff
@@ -83,12 +83,18 @@ def __hash__(self):
...
+        return [f"<y<{self.item}>> can be found in <r<{self.zone_override}>>"]
 
     def to_spoiler_log_text(self) -> str:
-        return f"{self.item} is on {self.location}"
+        if not self.zone_override:
+            return f"{self.item} is on {self.location}"
+        return f"{self.item} is in {self.zone_override}"
 
     def to_spoiler_log_json(self):
         return {"location": self.location, "item": self.item, "type": "item"}
```
lepelog, line 100: zone override needs to be added here

lepelog (changes requested):

lepelog:
In general you shouldn't make 2 completely separate changes in the same PR, opening a new one is usually better or we end up with `gui-redesign` again

YourAverageLink:
My main concern was that I wanted to make the zone item hints compatible with new cube handling, and I was unsure if I made a new branch based off of this one if that meant we'd just end up with all the same commits again, plus the new one

PR #183, opened 2022-05-16
Add long range cube west of ET hit and FS hook beetle skip

yannl35133:


lepelog (changes requested):
:+1: for hook beetle skip trick and I agree with the `Long Range Skyward Strike Jumpslash Cube Hits` trick, I would drop the `Cube Hits` from it though

Could you also change `LMF - Key Locked Room Bomb Throw` and `Temple of Time - Bomb Throw` to use the generic `Bomb Throws` trick? I would say that's in scope since you changed the Skyview one

PR #186, opened 2022-05-26
don't show first time textbox for treasures and rupees

lepelog:


lepelog:
build: https://nightly.link/lepelog/ssrando/workflows/build.yaml/remove-treasure-textbox/dist%20windows-latest

PR #188, opened 2022-06-11
add bomb flower and block way back from LMF bk room to prevent softlocks

lepelog:
Tested on dolphin, should be good to merge after a test on console


lepelog:
build: https://nightly.link/lepelog/ssrando/workflows/build.yaml/lmf-lock/dist%20windows-latest

PR #189, opened 2022-06-15
Path Hint Support

YourAverageLink:
(sorry, apparently rebasing locally duplicates all commits)


hints/hint_distribution.py
```diff
@@ -189,28 +193,20 @@ def start(self, logic: Logic, always_hints: list, sometimes_hints: list):
...
-                    continue
-
-            zone, specific_loc = Logic.split_location_name_by_zone(sots_loc)
-            self.sots_locations.append((zone, sots_loc, item))
+        self.sots_locations = self.loc_dict_filter(self.logic.sots_locations)
         self.rng.shuffle(self.sots_locations)
+        self.goals = list(self.logic.goal_locations.keys())
+        self.rng.shuffle(
+            self.goals
+        )  # shuffle the goal names that will be chosen in sequence when goal hints are placed to try to ensure one is placed for each goal
```
lepelog, line 201: Prefer comments above the code line, black is kinda stupid about formatting them otherwise

hints/hint_distribution.py
```diff
@@ -189,28 +193,20 @@ def start(self, logic: Logic, always_hints: list, sometimes_hints: list):
...
+        self.goal_locations = list(
+            [
+                (self.loc_dict_filter(self.logic.goal_locations[goal_name]))
+                for goal_name in self.goals
+            ]
+        )  # create corresponding list of shuffled goal items
```
lepelog, line 207:
```suggestion
        # create corresponding list of shuffled goal items
        self.goal_locations = [
            (self.loc_dict_filter(self.logic.goal_locations[goal_name]))
            for goal_name in self.goals
        ]
```
list([...]) is redundant, only [...] works the same way

logic/logic.py
```diff
@@ -1446,15 +1451,31 @@ def get_sots_locations(self):
...
+                            [loc], self.macros[POST_GOAL_LOCS[goal]]
+                        ):
+                            path_goals[goal][loc] = item
+        if not self.sots_locations or not self.goal_locations:
+            self.sots_locations, self.goal_locations = sots_items, path_goals
+            return (
+                sots_items,
+                path_goals,
+            )  # return only the first time this method is called
+        self.sots_locations, self.goal_locations = sots_items, path_goals
```
lepelog, line 1475: Why only return something the first time this method is called? That doesn't really make sense to me. If the goal was to cache `sots_locations` and `goal_locations` you should check first if both have been calculated and then return them, and only if it has not been calculated yet do the actual calculation.

But since you access `sots_locations` and `goal_locations` directly, maybe rename this method to `calculate_sots_path_locations` and have it return nothing

YourAverageLink: Actually in retrospect I think I may have misunderstood what he said. I think maybe he wanted to check if the cache exists at the start of the method and return it if so, and if not calculate it and then return

PR #192, opened 2022-06-19
Fix items shuffled with any restrictions being unable to be hinted

cjs8487:
**This is a critical S2 fix**

This fixes a bug that causes items that have restricted placements to be unable to be hinted by any type of hint besides always. Currently this only affects maps, small keys, and big keys, however it can be expanded for any other items that my be placed under restrictions in the future. This does not impact fully pre-randomized items such as swords at the end of dungeons in the sword reward mode.

This does have a slight side effect - if for whatever reason keys a fully pre-randomized manually by a code edit, they are still eligible to be hinted, which is contrary to the expected behavior. However, there is no way to tell the difference between an item that was pre-randomized directly and one that placed in the early randomization passes

### DIFF
--- a/hints/hint_distribution.py
+++ b/hints/hint_distribution.py
@@ -186,56 +186,22 @@ class HintDistribution:
         self.sometimes_hints = sometimes_hints
 
         # ensure prerandomized locations cannot be hinted
-        self.hinted_locations.extend(self.logic.prerandomization_item_locations.keys())
-
-        # filter out prerandomized keys
-        to_remove = []
-        for loc in self.hinted_locations:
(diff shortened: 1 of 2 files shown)


lepelog:
While I don't see a reason this doesn't work, 2 suggestions:
- make a function to check if an item is a dungeon item to be restricted (cause this code is at least duplicated twice, maybe more?)
- Instead of working with a temp list, I would suggest: `self.hinted_locations.extend((item for item in self.logic.prerandomization_item_locations.keys() if not self.logic.is_dungeon_restricted_item(item)))`

lepelog (approved):

PR #194, opened 2022-06-27
Add additional sometimes hints

cjs8487:
Adds the following sometimes hints
- Great Tree
- SV Dig Spot
- Turf Slide
- Slope Boulder Dig Spot (2x20)
- ET West Wing
- Fire Node Shortcut
- Lightning Node Ending
- Floria Chest
- AC East Wing
- SSH Brig Key

### DIFF
--- a/checks.yaml
+++ b/checks.yaml
@@ -651,7 +651,7 @@ Faron Woods - Chest inside Great Tree:
   Paths:
     - stage/F100_1/r0/l0/TBox/84
   hint: sometimes
-  text: the <r<hollow>> tree holds
+  text: the <r<hollow tree>> holds
 
 Lake Floria - Lake Floria Chest:
   original item: Goddess Plume
@@ -750,7 +750,6 @@ Mogma Turf - Chest behind Bombable Wall at Entrance:
 Mogma Turf - Sand Slide Chest:
   original item: Eldin Ore
   type: eldin, miscellaneous
-  hint: sometimes
   text: <r<behind>> a slide mogmas buried
   Paths:
     - stage/F210/r0/l0/TBox/68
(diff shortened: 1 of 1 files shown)


checks.yaml
```diff
@@ -650,10 +650,14 @@ Faron Woods - Chest inside Great Tree:
   type: faron, miscellaneous
   Paths:
     - stage/F100_1/r0/l0/TBox/84
+  hint: sometimes
+  text: the <r<hollow>> tree holds
```
lepelog, line 654: The coloring pattern seems inconsistent to me, sometimes it spans everything but the verb at the end, sometimes it just highlights a seemingly unrelated word.

I prefer highlighting all the important parts, so `hollow tree` in this case.

Also imo, we can just use `great tree` here.

Unless I'm the only one complaining, we could first get the new sometimes hints in without cryptic messages and add them later.

checks.yaml
```diff
@@ -1143,6 +1161,8 @@ Lanayru Mining Facility - Goddess Harp:
 Ancient Cistern - Chest in East Part:
   original item: AC Small Key
   type: faron, dungeon
+  hint: sometimes
+  text: the <r<lone west chest in the Ancient Cistern>> reveals
```
lepelog, line 1165: > Chest in East Part
> lone west chest

:thinking:

lepelog (changes requested):
mostly agree with the added hints, I've voiced my concerns about a few locations in discord already so I won't repeat it here

checks.yaml
YourAverageLink: This is a bit of a nitpick, but i disagree with the use of "reveals" here too

YourAverageLink:
The lightning node sometimes hint is missing a containing verb (like holds, contains, etc.) after its description

lepelog (approved):

PR #195, opened 2022-06-27
don't hint locations in barren areas

lepelog:


hints/hint_distribution.py
```diff
@@ -134,6 +142,9 @@ def start(self, logic: Logic, always_hints: list, sometimes_hints: list):
         self.rng = logic.rando.rng
         self.logic = logic
 
+        # don't allow junk hint copies
+        self.distribution["junk"]["copies"] = 1
```
cjs8487, line 146: Is this really a big deal? I hardly think someone would do this accidentally

hints/hint_distribution.py
```diff
@@ -525,6 +490,8 @@ def _create_random_hint(self):
                 for loc in self.logic.done_item_locations
                 if not loc in self.hinted_locations
                 and not loc in self.logic.prerandomization_item_locations
+                and Logic.split_location_name_by_zone(loc)[0]
```
cjs8487, line 493: For normal item hints this is impossible - the normal item list has no non-progress items in it which means a barren region can never have an item hint placed in it. However, that raises the question, if a distro adds a nonprogress item to the item list should that prevent barrens from being placed? Or should the item hint still be allowed to be placed? Or neither?

hints/hint_distribution.py
lepelog: `copies` was implicitly ignored before, so it was treated as if copies was always set to 1 and I didn't want to change existing behavior. I think junk hint copies don't make sense but they're also not a problem. If you're fine with changing existing behavior, I'll remove this line

hints/hint_distribution.py
cjs8487: oh thanks GitHub (fwiw it also confused YAL who is the who pointed this out to me lmao)

cjs8487 (approved):
lgtm for fixing the issue with random hints. I do think the issue with sometimes hints is probably worth revisiting in the future if we can solve hint balancing a bit more, but that's a discussion for later

PR #198, opened 2022-07-04
Add main node option for LMF

YourAverageLink:
Adds option for starting with each of the three subnodes activated, but still needing to activate the main generator to actually raise LMF.


lepelog (approved):
still needs some test runs

PR #203, opened 2022-07-07
Chest Sizes Match Contents option (mostly by CJ)

YourAverageLink:
Currently the UI looks weird and appearing chests have to be forced big. I recommend (if it gets merged at all) holding off on merging until after Season 2 is frozen.


CovenEsme:
Now that CDMC has been merged, does this need to remain open or can it get archived?

PR #208, opened 2022-07-20
Lower goddess wall requirement

lepelog:
- Changes the requirement for all goddess walls to Harp and BotG (removes scale, and mogma text flag for the sealed grounds one)
- Fixes the butterflies not spawning for the sealed grounds wall and for trials


Peppernicus2000:
Reviewed it and it works, also fixed the errors

PR #212, opened 2022-08-13
fix nonprogress calculation

lepelog:
refactors how nonprogress items are calculated, which now rightfully changes pouches to nonprogress if you start with one


Peppernicus2000:
With the new logic comming this PR is irrelevant "Quote Lepelog"

PR #214, opened 2022-08-13
Add Option to Randomize Triforces

YourAverageLink:
The Triforces of Courage, Wisdom, and Power may be randomized. If Triforce Required is enabled, the Triforce must be completed to open the door to Hylia's Realm. Triforces may be chosen to be vanilla, only in Sky Keep, or anywhere in the world. If the Triforce is not required, or they may appear anywhere, Sky Keep acts like an unrequired dungeon.


options.yaml
```diff
@@ -131,13 +131,23 @@
...
+  default: true
+  help: If activated, the three Triforces will be required to open the door to Hylia's Realm. If disabled, Sky Keep counts as an unrequired dungeon
+  ui: option_triforce_required
+- name: Triforce Shuffle
+  command: triforce-shuffle
+  type: singlechoice
+  bits: 2
+  choices:
+    - Vanilla
+    - Sky Keep
```
lepelog, line 146: Do we care about consistency with (Key|Map)sanity options and should use `Own Dungeon` here? I'm not really sure and I think it's fine how it is, just wanted to bring it up

patches.yaml
```diff
@@ -416,6 +416,72 @@ D003_8: # this change is purely to make sure D003_8 gets patched, so a perviousl
     layer: 6
     room: 0
     objtype: OBJ
+  - name: remove last triforce cutscene # this does mean if triforces are vanilla you have to leave yourself
+    type: objdelete
+    id: 0xFC06
+    layer: 0
+    room: 0
+    objtype: STAG
+  - name: Scene Change for Wisdom layer
```
lepelog, line 425: Scene changes can also be activated by sceneflags, so maybe just adding an `ScChang` with the right params could work?

YourAverageLink: Probably. I just copied the event that sends you out of the dungeon when the triforce is complete and adjusted its params to activate from the item sceneflags and connect to new entrypoints I made because I thought that was the simplest approach.

YourAverageLink: Update: I tried using `ScChang`s but that resulted in Link being sent to the silent realm state in sky keep. I think it should thus just be merged as it is.

options.yaml
YourAverageLink: I think Sky Keep is fine since they only ever appear in Sky Keep, unlike keys which exist in every dungeon (save for ET small keys)

patches.yaml
YourAverageLink: Works with nontrial patch.

logic/constants.py
```diff
@@ -28,6 +28,10 @@
         ("SS", "Sandship"),
         ("FS", "Fire Sanctuary"),
         ("SK", "Sky Keep"),
+        (
+            "Triforce",
+            "Sky Keep",
+        ),  # for the purposes of restricting triforces to sky keep
```
lepelog, line 34:
```suggestion
        # for the purposes of restricting triforces to sky keep
        ("Triforce", "Sky Keep"),
```
If you put the comment above the line it's commenting on instead of after it, the formatter won't break lines stupidly like that

lepelog (changes requested):

lepelog (approved):

PR #215, opened 2022-08-14
Rupeesanity

elijah-thomas774:
Adds Freestanding rupees that have a scene flag to the randomizer with an option.

current options: Vanilla / No quick Beetle / All

The current implementation of quick beetle is locked behind access to its materials. Will be changed after Raid's PR is verified.

When set to vanilla, the checks act as if they were not touched (no text box)


logic/logic.py
```diff
@@ -204,6 +206,16 @@ def __init__(self, rando):
                 else:
                     self.racemode_ban_location(shop_check)
 
+        if self.rando.options["rupeesanity"] != "All":
+            for rupee_check in RUPEE_CHECKS:
+                if self.rando.options["rupeesanity"] == "Vanilla":
+                    orig_item = self.item_locations[rupee_check]["original item"]
+                    self.set_prerandomization_item_location(rupee_check, orig_item)
+                else:
```
lepelog, line 214: I think a comment which rupeesanity setting is in the else branch (the one without upgraded beetle) would make it more clear why those checks are forced to be vanilla.

Also I think it would be better to have the affected checks in a list in `logic/constants.py` since I'm not a fan of relying on parts of the name (which could change in the future)

gamepatches.py
```diff
@@ -933,13 +934,21 @@ def load_base_patches(self):
         self.startitemflags = self.patches["global"]["startitems"]
 
         # patches from randomizing items
+        temp_item_locations = copy.deepcopy(self.placement_file.item_locations)
+        if self.placement_file.options["rupeesanity"] != "All":
+            for rupee_check in RUPEE_CHECKS:
+                if self.rando.options["rupeesanity"] == "Vanilla":
+                    temp_item_locations.pop(rupee_check)
+                elif "Sea Pillar" in rupee_check or "Entrance Crown" in rupee_check:
+                    temp_item_locations(rupee_check)
```
lepelog, line 943: Is this supposed to have a `.pop` as well?

lepelog (changes requested):

logic/logic.py
elijah-thomas774: This is Temporary until Progressive Items are added in. Once that gets merged in, itll need to change to be linked to the progressive tools option

lepelog (approved):

PR #220, opened 2022-09-18
Boss key hints

Muzugalium:
Added hints that point to the location of the boss key for required dungeons

### DIFF
--- a/hints/hint_distribution.py
+++ b/hints/hint_distribution.py
@@ -528,7 +528,7 @@ class HintDistribution:
             self.logic.done_item_locations[loc],
             True,
             self.logic.item_locations[loc].get("text"),
-            "boss key",
+            "boss_key",
         )
 
     def _create_junk_hint(self):
--- a/hints/hint_distribution.py
+++ b/hints/hint_distribution.py
@@ -115,6 +115,7 @@ class HintDistribution:
             "item": self._create_item_hint,
             "random": self._create_random_hint,
             "junk": self._create_junk_hint,
+            "bk": self._create_bk_hint,
         }
 
     def read_from_file(self, f):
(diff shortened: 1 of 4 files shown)


hints/hint_distribution.py
```diff
@@ -505,6 +516,21 @@ def _create_random_hint(self):
...
+        loc = self.required_boss_key_locations.pop()
+        if loc in self.hinted_locations:
+            return self._create_bk_hint()
+        self.hinted_locations.append(loc)
+        return LocationGossipStoneHint(
+            loc,
+            self.logic.done_item_locations[loc],
+            True,
+            self.logic.item_locations[loc].get("text"),
+            "boss key",
```
yannl35133, line 531: This string goes into JSON objects, so it's better for spaces to be replaced by `_`

lepelog (changes requested):
- boss key hints have to be added to the changelog
- currently these changes do nothing so bk hints would need to be added to some existing hint distribution or a new one
- if a BK is on LMF BK chest or 3rd small key, that's always hinted and not counted towards the fixed number of BK hints, I guess that's intended?

YourAverageLink:
> - boss key hints have to be added to the changelog
> - currently these changes do nothing so bk hints would need to be added to some existing hint distribution or a new one
> - if a BK is on LMF BK chest or 3rd small key, that's always hinted and not counted towards the fixed number of BK hints, I guess that's intended?

The last two points are likely meant to be solved by new hint distributions. The dungeon always hints would be downgraded to sometimes hints or removed, since the BK hint sort of takes the role those hints used to fill.

Muzugalium:
For the last two points, yep exactly as Yal said

lepelog (approved):
I still think it's weird to add an unused feature (since it's not used in any distro) but I guess it's fine

PR #223, opened 2022-09-26
Dusk relic randomizing

Peppernicus2000:
Randomizes 5 dusk relics per Trials so 15 new checks. Changes Trial region hints when option is set (currently default until UI got the option), by allowing barren and sots hints and changeing trial reward hints to sometimes hints. Also works with Trial object shuffle, although for eldin it will only move the items around in room 2 (still like 20 to 30 locations it can be)


YourAverageLink:
Do you intend to keep the song hints option only referring to the reward check? Or should it be barren vs nonbarren for basic hints for all checks inside (likewise for barren vs nonbarren vs sots for advanced)?

Peppernicus2000:
> 

currently you both have a barren/nonbarren/sots hint for the trial realms and i kept the reward hints, but moved them to sometimes hints. so the spefic saying will only talk about the actually reward from finishing the trial. I didn't think much about the actually implementations so i am happy to change things if that makes it feel better, but for that i would need some people to test it

gamepatches.py
```diff
@@ -1933,6 +1954,65 @@ def shuffle_trial_objects(self):
...
+                and not self.placement_file.options["shuffle-trial-objects"] == "None"
+            ):
+                for (id, room) in list(relic_list):
+                    if room != 2:
+                        relic_list.remove((id, room))
+            locs.extend(
+                random.sample(
+                    relic_list, self.placement_file.options["trial-treasure-amount"]
+                )
+            )
```
yannl35133, line 2000:
```suggestion
            rng = random.Random(self.placement_file.trial_object_seed)
            locs.extend(
                rng.sample(
                    relic_list, self.placement_file.options["trial-treasure-amount"]
                )
            )
```

YourAverageLink:
Replaced by Pull Request #413. I think this can be closed now.

PR #224, opened 2022-09-27
More text/cutscene refinements

YourAverageLink:


eventpatches.yaml
```diff
@@ -1005,6 +1005,11 @@
     cases:
       - Rerandomize Treasure
       - -1
+  - name: Skip Fledge 3
+    type: flowpatch
+    index: 268 # after giving fledge the second stamina potion
```
CovenEsme, line 1010: Currently, this is only after giving the 1st stamina potion.
Also, cutting out more of the text would be preferable. There's still quite a lot of it to mash through - even with just the one conversation.

eventpatches.yaml
```diff
@@ -1134,6 +1139,11 @@
       type: type1
       next: Kina Quest Switch
       param4: New Kina Text Option
+  - name: Shorten Kina Text
```
CovenEsme, line 1142: Works well.
Would be great if Eagus' and Pumm's text was also shortened to match.

CovenEsme (changes requested):
Everything works except fledge gives the item after 1 potion instead of 2.
Would be cool if more of the kina and fledge text was skipped but definitely not necessary.

eventpatches.yaml
```diff
@@ -1184,6 +1194,11 @@
     index: 7
     flow:
       next: 21 # never talk about triforce or song of the hero
```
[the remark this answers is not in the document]
YourAverageLink: I think I was trying to skip a long text sequence taht was on 21

PR #235, opened 2022-10-07
Presets

cjs8487:
Adds the ability for users to create, update, and manage their own presets, as well as a system for default presets distributed with the randomizer program. Presets are stored in JSON format (like saved settings) and could potentially be moved to remote storage in the future.

The defaults included with this are all 6 S2 modes, and a new beginner friendly preset.

Notes:
- Default presets are protected. They cannot be saved or deleted.
- Presets cannot share a name
- Order in the preset file matters, as that is the order they will be displayed. For user presets this should always be the order of creation
- Selecting a new preset **does not** overwrite the currently selected options until the load button is pressed
- Default presets are always loaded first, and then there is a divider in the selection list


gui/default_presets.json
```diff
@@ -0,0 +1,9 @@
...
...-thunderhead": "Ballad", "starting-sword": "Goddess Sword", "required-dungeon-count": 3, "imp2-skip": true, "empty-unrequired-dungeons": true, "banned-types": ["peatrice", "goddess", "medium", "expensive"], "triforce-required": false, "triforce-shuffle": "Vanilla", "seed": -1, "hero-mode": true, "randomize-entrances": "None", "randomize-trials": false, "start-with-pouch": true, "no-spoiler-log": true, "max-batreaux-reward": 50, "shop-mode": "Randomized", "rupoor-mode": "Added", "rupeesanity": "Vanilla", "gondo-upgrades": false, "got-start": "Lowered", "got-sword-requirement": "Master Sword", "got-dungeon-requirement": "Required", "open-lmf": "Open", "skip-horde": false, "skip-g3": false, "skip-demise": false, "map-mode": "Own Dungeon - Restricted", "small-key-mode": "Lanayru Caves Key Only", "boss-key-mode": "Own Dungeon", "logic-mode": "BiTless", "enabled-tricks-bitless": [], "enabled-tricks-glitched": [], "song-hints": "None", "fix-bit-crashes": false, "impa-sot-hint": false, "sword-dungeon-reward": false, "open-et": true, "demise-count": 1, "cutoff-gameover-music": false, "cube-sots": false, "precise-item": false, "shuffle-trial-objects": "None", "hint-distribution": "S2 - 3D"},
```
YourAverageLink, line 5:
```suggestion
    "S2 - 3D Open": {"starting-tablet-count": 3, "open-thunderhead": "Open", "starting-sword": "Goddess Sword", "required-dungeon-count": 3, "imp2-skip": true, "empty-unrequired-dungeons": true, "banned-types": ["peatrice", "goddess", "medium", "expensive"], "triforce-required": false, "triforce-shuffle": "Vanilla", "seed": -1, "hero-mode": true, "randomize-entrances": "None", "randomize-trials": false, "start-with-pouch": true, "no-spoiler-log": true, "max-batreaux-reward": 50, "shop-mode": "Randomized", "rupoor-mode": "Added", "rupeesanity": "Vanilla", "gondo-upgrades": false, "got-start": "Lowered", "got-sword-requirement": "Master Sword", "got-dungeon-requirement": "Required", "open-lmf": "Open", "skip-horde": false, "skip-g3": false, "skip-demise": false, "map-mode": "Own Dungeon - Restricted", "small-key-mode": "Lanayru Caves Key Only", "boss-key-mode": "Own Dungeon", "logic-mode": "BiTless", "enabled-tricks-bitless": [], "enabled-tricks-glitched": [], "song-hints": "None", "fix-bit-crashes": false, "impa-sot-hint": false, "sword-dungeon-reward": false, "open-et": true, "demise-count": 1, "cutoff-gameover-music": false, "cube-sots": false, "precise-item": false, "shuffle-trial-objects": "None", "hint-distribution": "S2 - 3D"},
```
Also, while not specific to Open, you forgot to remove cutoff gmaeover music from the saved settings

gui/randogui.py
```diff
@@ -531,6 +563,76 @@ def disable_trick(self):
         self.ui.disabled_tricks.model().sort(0)
         self.update_settings()
 
+    def load_preset(self):
+        preset = self.ui.presets_list.currentText()
+        # prevent loading the new preset option
```
lepelog, line 568: I don't know anything about QT, but how hard/how much work would it be to just disable the buttons when you're not supposed to click them? I think that would be nicer to use

gui/default_presets.json
```diff
@@ -0,0 +1,9 @@
...
... "open-thunderhead": "Ballad", "starting-sword": "Goddess Sword", "required-dungeon-count": 3, "imp2-skip": true, "empty-unrequired-dungeons": true, "banned-types": ["minigame", "peatrice", "goddess", "medium", "expensive"], "triforce-required": false, "triforce-shuffle": "Vanilla", "seed": -1, "hero-mode": true, "randomize-entrances": "None", "randomize-trials": false, "start-with-pouch": true, "no-spoiler-log": true, "max-batreaux-reward": 50, "shop-mode": "Randomized", "rupoor-mode": "Added", "rupeesanity": "Vanilla", "gondo-upgrades": false, "got-start": "Lowered", "got-sword-requirement": "Master Sword", "got-dungeon-requirement": "Required", "open-lmf": "Open", "skip-horde": false, "skip-g3": false, "skip-demise": false, "map-mode": "Own Dungeon - Restricted", "small-key-mode": "Anywhere", "boss-key-mode": "Anywhere", "logic-mode": "BiTless", "enabled-tricks-bitless": [], "enabled-tricks-glitched": [], "song-hints": "None", "fix-bit-crashes": false, "impa-sot-hint": false, "sword-dungeon-reward": false, "open-et": true, "demise-count": 1, "cutoff-gameover-music": false, "cube-sots": false, "precise-item": false, "shuffle-trial-objects": "None", "hint-distribution": "S2 - 3D"}
```
lepelog, line 8: Remove `no-spoiler-log` from all of them

gui/randogui.py
```diff
@@ -113,6 +116,35 @@ def __init__(self, options: Options):
         self.ui.enable_trick.clicked.connect(self.enable_trick)
         self.ui.disable_trick.clicked.connect(self.disable_trick)
 
+        # setup presets
+        self.default_presets = {}
+        self.user_presets = {}
+        self.ui.presets_list.addItem("[New Preset]")
+        sep_idx = 1
+        with open("gui/default_presets.json") as f:
```
lepelog, line 124:
```suggestion
        with (RANDO_ROOT_PATH / 'gui' / 'default_presets.json').open() as f:
```

otherwise the exe breaks

lepelog (changes requested):
KolokChamp

gui/randogui.py
cjs8487: I left the checks in just in case there's a way around QTs disable, but the buttons now enable/disable as appropriate

gui/randogui.py
```diff
@@ -531,6 +563,76 @@ def disable_trick(self):
...
+                    self.error_msg = QErrorMessage()
+                    self.error_msg.showMessage("Cannot have duplicate preset names")
+                    return
+                else:
+                    preset = name
+                    self.ui.presets_list.addItem(preset)
+                    self.ui.presets_list.setCurrentText(preset)
+        self.user_presets[preset] = self.options.to_dict(
+            [
+                "dry-run",
```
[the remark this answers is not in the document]
cjs8487: I combined both ideas - all non-permalink options should be excluded from presets (at least as of now), however there are some edge case options, such as `no-spoiler-log`, where the option is in the permalink but should be excluded from presets as well. This setup will give us the flexibility to handle both cases, and we can always fall back to a full exclusionary version in the future if the need arises

gui/randogui.py
```diff
@@ -113,6 +116,37 @@ def __init__(self, options: Options):
...
+                for preset in load_default_presets:
+                    self.ui.presets_list.addItem(preset)
+                    self.default_presets[preset] = load_default_presets[preset]
+                    sep_idx += 1
+            except Exception as e:
+                print("couldn't load default presets")
+        self.ui.presets_list.insertSeparator(sep_idx)
+        self.user_presets_path = "presets.txt"
+        if os.path.isfile(self.user_presets_path):
+            with open(self.user_presets_path) as f:
```
yannl35133, line 136: Is this a purely relative file opening ? Shouldn't you at least start from `RANDO_ROOT_PATH` ?

options.py
```diff
@@ -204,9 +204,15 @@ def update_from_permalink(self, permalink: str):
                 raise Exception(f'unknown type: {option["type"]}')
             self.set_option(option_name, value)
 
-    def to_dict(self):
+    def to_dict(self, include_nonperma=True, exclude=[]):
         opts = self.options.copy()
         for option_name, option in OPTIONS.items():
+            if not include_nonperma and "permalink" in option and not option["permalink"]:
```
yannl35133, line 210:
```suggestion
            if exclude_nonperma and option.get("permalink", True) is False:
```
Of course the `exclude_nonperma` would have to be reflected at all (1) call sites

gui/randogui.py
```diff
@@ -113,6 +116,37 @@ def __init__(self, options: Options):
         self.ui.enable_trick.clicked.connect(self.enable_trick)
         self.ui.disable_trick.clicked.connect(self.disable_trick)
 
+        # setup presets
+        self.default_presets = {}
+        self.user_presets = {}
+        self.ui.presets_list.addItem(new_preset)
+        sep_idx = 1
+        with (RANDO_ROOT_PATH / 'gui' / 'default_presets.json').open('r') as f:
+            try:
```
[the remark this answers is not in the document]
cjs8487: This is hardly the place for such a refactor. If we decide that is the path forward it should developed and tested as an isolated set of changes

gui/randogui.py
cjs8487: It’s purely relative. It will always be at the root of the source or in the same directory as the exe. It’s the same as the settings file

lepelog (approved):

PR #236, opened 2022-10-08
Add Flooded Faron region

CovenEsme:
Adds Flooded Faron Woods as an accessible region in rando.

Working:
* Entry and exit to and from Flooded Faron
* Faron SotH check (collecting all tadtones)
* Alternative logic for Owlan Crystal's
  * Access to Flooded Faron as an alternative to bomb bag

Not Working:
* Machi, Lopsa and Erla become saved on entering Flooded Faron


CovenEsme:
Added a fix (removing fi text) that started the SotH story quest when on upper great tree layer 5 (owlan crystals started) and have completed tadtones.

Fixed issue where you couldn't leave Flooded Faron Woods after starting but not completing tadtones.

hints/hint_distribution.py
```diff
@@ -235,7 +235,7 @@ def start(self, logic: Logic, always_hints: list, sometimes_hints: list):
 
         region_barren, nonprogress = self.logic.get_barren_regions()
         for zone in region_barren:
-            if "Silent Realm" in zone:
+            if "Silent Realm" in zone or "Flooded Faron Woods":
```
yannl35133, line 238:
```suggestion
            if "Silent Realm" in zone or zone == "Flooded Faron Woods":
```
Currently every region is skipped since `"Flooded Faron Woods"` casts to true

CovenEsme: Fixed :p

olohr: python moment

PR #237, opened 2022-10-08
remove unnecessary layeroverrides from behind the temple

lepelog:


lepelog:
Currently, the storyflags for the other layers are never set, but that will probably change with the tadtones PR, so this is preparation for that

Peppernicus2000:
Yes it looks like this was one of the first layeroverwrites that was done were i might have though, or that might have even been the case then, you needed to write down all the layers to have it work. If the other layers get reached that should be fine looking at the object map so we shouldn't need to change anything just remove the unnesseary

PR #239, opened 2022-10-09
try to make bulk mode work on bad operating systems

lepelog:


CovenEsme:
I think I left feedback in discord a while back but this works great on windows.

PR #241, opened 2022-10-09
Add stamina potion for logical Lanayru access

CovenEsme:


yannl35133 (commented):
I would have a `Endurance Potion` macro and would use it both for Fledge and in Lanayru, so that the "LMF Open" part of the requirement is better explained

CovenEsme:
An `Unlocked Endurance Potion` marco already existed. Changed to used that.

Peppernicus2000:
Can you add a changelog change so people know that the logic changed when useing open LMF

PR #248, opened 2022-10-14
Add heart container sword reward option

YourAverageLink:
Changes sword reward to singlechoice, adds option to force swords on the vanilla heart container check


lepelog (approved):
should be good once the merge conflicts are resolved

PR #253, opened 2022-10-20
Remove non-s2 presets

CovenEsme:


lepelog:
I'm not a fan of removing these presets, not sure what the motivation is, I would rather close this PR

CovenEsme:
Can be closed. PR exists to fix CJ's issues with other people creating presets. Awaiting a standard preset for 1.3

PR #254, opened 2022-10-22
Changes in new logic that pollute the diff

yannl35133:


logic/constants.py
```diff
@@ -1,50 +1,87 @@
...
+        (LMF, LMF),
+        (AC, AC),
+        (SSH, SSH),
+        (FS, FS),
+        (SK, SK),
         # for the purposes of restricting triforces to sky keep
-        ("Triforce", "Sky Keep"),
-        ("LanayruCaves", "Lanayru Caves"),  # "short name" doesn't allow space
+        ("Triforce", SK),
+        ("Lanayru Caves", "Lanayru Caves"),  # "short name" doesn't allow space
```
lepelog, line 63: Is this comment still accurate? If not, it should be removed

lepelog (changes requested):
I think having the full dungeon name in the key items is good, one small nitpick otherwise looks good

PR #255, opened 2022-10-22
Get all new Dowsing Options after getting Whitesword

Peppernicus2000:
Option currently set as default needs ui implementation
- ItemGet for whitesword changed when haveing Option set 
- Starting sword equal or higher then whitesword now actually gives the extra dowsing slots
- If option is set also sets dowsings when haveing Starting sword equal or higher then whitesword

### DIFF
--- a/gamepatches.py
+++ b/gamepatches.py
@@ -1491,6 +1491,13 @@ class GamePatcher:
         )
         for i in range(start_sword_count):
             self.startstoryflags.append(PROGRESSIVE_SWORD_STORYFLAGS[i])
+        if start_sword_count > 3:
+            self.startstoryflags.append(583) # 4 extra Dowsing slots
+            if self.placement_file.options["dowsing-after-whitesword"]:
+                self.startstoryflags.append(102) # treasure Dowsing
+                self.startstoryflags.append(104) # Crystal Dowsing
+                self.startstoryflags.append(105) # Rupee Dowsing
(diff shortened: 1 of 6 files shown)


gamepatches.py
```diff
@@ -1501,6 +1501,13 @@ def add_startitem_patches(self):
         )
         for i in range(start_sword_count):
             self.startstoryflags.append(PROGRESSIVE_SWORD_STORYFLAGS[i])
+        if start_sword_count > 3:
+            self.startstoryflags.append(583)  # 4 extra Dowsing slots
+            if self.placement_file.options["dowsing-after-whitesword"]:
+                self.startstoryflags.append(102)  # treasure Dowsing
+                self.startstoryflags.append(104)  # Crystal Dowsing
+                self.startstoryflags.append(105)  # Rupee Dowsing
+                self.startstoryflags.append(110)  # Goddess Cube Dowsing
```
YourAverageLink, line 1510: If you move these lines out of the start_sword_count check block, you won't need the other flow patches you made in eventpatches afaik; those extra dowsing options won't be available if you don't have the full wheel anyway (unlocked by whitesword)

Peppernicus2000: you can still select them even if you don't see them, so sadly that is not an option. basicly the eventpatches are for the case that you get the whitesword later in the run and the code in the start sword block for when you start with whitesword or higher

PR #259, opened 2022-10-23
Patch actor params and refactor item stagepatches to use normal patches

lepelog:
- makes it easier to just patch certain parts of parameters for objects according to `actor_params.yaml`
- don't special case item patches anymore, they are just normal patches now
- add `query` to make finding object entries more flexible

This slightly changes the order patches are applied, in the past they always followed `objadd` -> `objpatch` etc, but now they are strictly in the order they were defined in


lepelog:
Closing because I don't like this idea anymore

PR #270, opened 2022-11-05
Add OHKO Mode

YourAverageLink:
April Fools came early/late

Adds OHKO mode where all damage one hit kills Link. Logic accounts for potentially required damage. Currently does not account for Guardian Potion + in logic.


cjs8487 (commented):
Should OHKO off be added to all bomb combat? I know you can get around getting bokos to hit the bomb by dropping from above, but that's not always a strat that's available. I'm not sure how every enemy interacts with placed, lit bombs

lepelog:
I think it makes sense to instead make this a "damage multiplier" option and to have "Can survive X damage" macros (might make sense to wait for new logic for that)

lepelog:
closing since https://github.com/ssrando/ssrando/pull/296 is merged

PR #271, opened 2022-11-08
Add hints on NPCs for some checks

YourAverageLink:
Adds NPC hints for
- Faron SotH
- Kina
- Owlan
- Levias

Also fixes song hint mapping bug & puts song/npc hints on the spoiler log.

### DIFF
--- a/SpoilerLog.py
+++ b/SpoilerLog.py
@@ -7,7 +7,16 @@ from version import VERSION
 from logic.constants import DUNGEON_GOALS, DEMISE
 
 
-def write(file: TextIO, options: Options, logic, hints, sots_locations, hash):
+def write(
+    file: TextIO,
+    options: Options,
+    logic,
+    hints,
+    song_hints,
+    npc_hints,
+    sots_locations,
+    hash,
+):
     write_header(file, options, hash)
 
     if options["no-spoiler-log"]:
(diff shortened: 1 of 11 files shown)


CovenEsme:
Not sure how to deal with the conflicts here so I've left this PR out of the `beta-features` branch until they get resolved

CovenEsme:
Is this feature still desired? If so, it could do with updating so it can get tested and merged

cjs8487:
I believe we decided that we wanted the feature, but that there were implementation problems, along with some debate about the specifics of the feature. From what I recall, there were two major things that were blocking this:
1. How exactly we wanted triforce hints to work, should it be one static piece chosen randomly at generation, or should you be able to dynamically receive some number of the hints based on what you have collected? This also begs the further question of how it would interact with TFH, but that’s a bridge that we can cross when TFH becomes a reality
2. Turning it into a conditional multi so that it functions similarly to the equivalent functionality in OoTR, and so that we can maintain current functionality without making the Impa hint a special case/separate option.

In the interim, I’ve actually come up with a few other notes on this that I never got around to writing up anywhere and eventually forgot about this.
- We could likely rename this to “Miscellaneous Hints.” Even though this currently only impacts NPCs, I could see merit in adding some variations of this style of hint on actors that aren’t NPCs (such as signs).
- These hints should probably not remove the hint status of locations that it hints, and instead rely on deduplication to prevent potential bugs when it is disabled
- Using the conditional multi implementation should eventually allow us to give better and more specific help text for each option, which would mean the help text for the option as a whole doesn’t have to worry about specific cases (such as the fact that levias is the only one that becomes an always hint)
- This doesn’t necessarily need to be part of this particular PR, but Batreaux rewards and Beedle’s shop items would also make great candidates for this style of hint (on an NPC or otherwise)

YourAverageLink:
Was clearing out some old branches on my fork. If we want this back again, it'll probably be after #610 since it'll warrant a redo anyway

PR #272, opened 2022-11-10
Add Support for Custom Hint Distributions

cjs8487:
This adds support for user provided custom hint distributions.  Users are assumed to have functional knowledge of how to create their own distributions when using this option. Error messages regarding invalid distributions are deliberately vague as there is no specific error checking on distributions (currently the only time a distribution can explicitly raise an error is when it un able to be read due to bad JSON). There can only be a single active custom distribution at a time, and it will always be sourced from `custom_hint_distribution.json`. The non-whitespace contents of the file are added to the randomizers hash process when this option is enabled.

Additionally a secondary error check has been added to the UI, allowing the randomizer's initialization process to raise errors and have them handled properly by the UI. This should also help catch some edge error cases that may pop up for end users and allow us to debug them more effectively.


ssrando.py
```diff
@@ -120,6 +120,15 @@ def _get_rando_hash(self):
         current_hash.update(str(self.seed).encode("ASCII"))
         current_hash.update(self.options.get_permalink().encode("ASCII"))
         current_hash.update(VERSION.encode("ASCII"))
+        if self.options["hint-distribution"] == "Custom":
+            if not os.path.exists("custom_hint_distribution.json"):
```
lepelog, line 124: I would refactor this and add `CUSTOM_HINT_DISTRIBUTION_PATH = Path("custom_hint_distribution.json")` to paths.py and then import that everywhere (that also allows to use `CUSTOM_HINT_DISTRIBUTION_PATH.open('r')` and `CUSTOM_HINT_DISTRIBUTION_PATH.exists()`)

ssrando.py
```diff
@@ -120,6 +120,15 @@ def _get_rando_hash(self):
...
+                raise Exception(
+                    "Custom hint distribution file not found. Make sure custom_hint_distribution.json exists at the same location as the randomizer"
+                )
+            with open("custom_hint_distribution.json") as f:
+                lines = f.readlines()
+                for line in lines:
+                    current_hash.update(line.strip(" \n").encode("ASCII"))
```
lepelog, line 131:
```suggestion
            with open("custom_hint_distribution.json") as f:
                normalized_json = json.dumps(json.load(f))
                current_hash.update(normalized_json.encode('ASCII'))
```
Checking a bit for how to hash dictionaries in python, it seems like most recomment just dumping to json and hashing that. This should also get rid of line ending problems

hints/hint_distribution.py
```diff
@@ -116,7 +116,12 @@ def __init__(self):
         }
 
     def read_from_file(self, f):
-        self._read_from_json(json.load(f))
+        try:
+            self._read_from_json(json.load(f))
+        except:
```
lepelog, line 121: I think it might be better to at least print the original exception

lepelog (changes requested):

YourAverageLink (commented):
Dunno if you want to given there's a new logic PR that addresses it and new logic is hopefully around the corner, but it would be nice if in this same PR, n-hints-per-stone functionality could be included while we're at it.

lepelog:
> 



> Dunno if you want to given there's a new logic PR that addresses it and new logic is hopefully around the corner, but it would be nice if in this same PR, n-hints-per-stone functionality could be included while we're at it.

This has nothing to do with that, let's not put a bunch of unrelated things into one PR

lepelog (approved):

lepelog:
can be merged after conflicts are resolved

PR #273, opened 2022-11-10
remove need to reload when light pillar is activated

lepelog:


yannl35133 (approved):
I tested this, it works. Could it be expanded to account for BotG and Thunderhead entrance, or would this be another thing completely ?

lepelog:
> I tested this, it works. Could it be expanded to account for BotG and Thunderhead entrance, or would this be another thing completely ?

That's a separate actor and unrelated to the actor modified in this patch. Further research would be needed, I think that actor is also more complicated although I haven't looked into it

PR #275, opened 2022-11-10
Remove first time textboxes 2: electric boogaloo

CovenEsme:
Removes first time textboxes for:
* Green, blue and red Rupees
* Heart flowers
* Arrow drops (singles and bundles)
* Bombs (5 and 10)
* Stamina Fruits
* Silent Realm Tears
* Silent Realm Light Fruits


lepelog (approved):
Since trials were tested, this should be good to go

PR #276, opened 2022-11-10
Add option to auto skip region intro cutscenes

CovenEsme:


CovenEsme:
I swear there's a better way to do this but I can't think of it off hand

PR #279, opened 2022-11-16
Edit options.yaml

goldendevil1711:
Some minor edits to the options in the UI to clarify and fix typographical or grammatical errors.

### DIFF
--- a/options.yaml
+++ b/options.yaml
@@ -174,7 +174,7 @@
     - All Dungeons
     - All Dungeons + Sky Keep
   default: None
-  help: "Shuffles entrances with one another. The Required Dungeons Separately option will shuggle only Required Dungeons with each other and Unrequired Dungeons with each other."
+  help: "Shuffles entrances with one another. The Required Dungeons Separately option will shuffle only Required Dungeons with each other and Unrequired Dungeons with each other."
   ui: option_randomize_entrances
 - name: Randomize Silent Realms
   command: randomize-trials
--- a/options.yaml
+++ b/options.yaml
@@ -187,7 +187,7 @@
   type: boolean
   default: false
   help: If activated, you will start with the adventure pouch unlocked. The remaining four
-    progressive pouche upgrades will still be shuffled.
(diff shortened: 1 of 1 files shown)


options.yaml
```diff
@@ -542,7 +544,7 @@
   min: 1
   max: 10
   bits: 4
-  help: How many demises you want to fight at the end of the game
+  help: How many Demises you want to fight at the end of the game.  More than 1 will cause significant lag to your game.
```
lepelog, line 547: 2 isn't too bad, iirc 3 was starting to lag but was still mostly fine and I haven't messed around with more on console (dolphin doesn't lag afaik)

options.yaml
```diff
@@ -60,7 +61,7 @@
     - True Master Sword
   default: Goddess Sword
   help: Select which sword to start with. The remaining upgrades will be shuffled into
-    the item pool
+    the item pool (based on other settings)
```
lepelog, line 64: I'm not sure about this addition. Since "based on other settings" basically means they are treated as progress items, I'm not sure if this actually clarifies things. Maybe `The remaining upgrades will be shuffled normally`?

goldendevil1711:
How about "The remaining upgrades will be shuffled normally (unless swords are set as dungeon rewards)"?

options.yaml
goldendevil1711: Fair enough.  Maybe you can delete that edit.

options.yaml
cjs8487: That’s still unclear imo because the rules for sword reward aren’t constant since they depend on how many dungeons are required and how many swords can be shuffled. I feel like just saying something similar to “the remaining upgrades will appear as normal” should be sufficient. There just isn’t an effective way to express this because of the semi-complex overlap in the options

lepelog:
This has a lot of conflicts, are you going to update this @goldendevil1711 ? It might be easier to close this and make a new PR, I also don't know if parts of this have already been implemented by other PRs

PR #280, opened 2022-11-16
General tweaks

CovenEsme:
* Updates and makes consistent the UI tooltips
* Adds the "song" banned type to Lanayru Soth and Faron Soth checks
* Combine Beedle's Shop options into a single drop-down
* Show previously hidden "Glitchless" logic
  * Basically BiTless but tricks cannot be enabled without switching to a different logic mode (behaviour previously defined by CJ)
  * Means that the default isn't "BiTless" anymore (which has been reported to be confusing to new players)
* Prevent "dungeon" banned type from being disabled in the UI (seeds cannot generate with it disabled)


CovenEsme:
Obsoleted by CJ's GUI changes.

PR #282, opened 2022-11-20
Exclude Individual Locations

cjs8487:
UI isn't currently the greatest for this but it will be addressed with the remainder of the UI changes

### DIFF
--- a/gui/randogui.py
+++ b/gui/randogui.py
@@ -117,6 +117,8 @@ class RandoGUI(QMainWindow):
         self.ui.enable_trick.clicked.connect(self.enable_trick)
         self.ui.disable_trick.clicked.connect(self.disable_trick)
 
+        print(OPTIONS["excluded-locations"]["choices"])
+
         # setup exlcuded locations
         self.excluded_locations_model = QStringListModel()
         self.excluded_locations_model.setStringList(
--- a/gui/randogui.py
+++ b/gui/randogui.py
@@ -128,6 +128,8 @@ class RandoGUI(QMainWindow):
         )
         self.ui.excluded_locations.setModel(self.excluded_locations_model)
         self.ui.included_locations.setModel(self.included_locations_model)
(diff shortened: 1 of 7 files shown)


gui/randogui.py
```diff
@@ -431,8 +440,26 @@ def update_ui_for_settings(self):
...
+            current_settings["excluded-locations"]
+        )
+        self.included_locations_model.setStringList(
+            [
+                choice
+                for choice in OPTIONS["excluded-locations"]["choices"]
+                if choice not in current_settings["excluded-locations"]
+            ]
+        )
+        self.excluded_locations_model.sort(0)
```
lepelog, line 458: Does it make sense to just default sort them? I think it makes more sense to have them sorted by their order in `checks.yaml`

options.py
```diff
@@ -9,6 +9,10 @@
     OPTIONS_LIST = yaml.safe_load(f)
 
 OPTIONS = OrderedDict((option["command"], option) for option in OPTIONS_LIST)
+with (RANDO_ROOT_PATH / "checks.yaml").open("r") as f:
```
lepelog, line 12: You can use `read_yaml_file_cached` here

lepelog (changes requested):
I think it would make sense to have the labels `Include Locations` and `Exclude Locations`, otherwise it's ambiguous

gui/randogui.py
cjs8487: The current order in `checks.yaml` is semi-arbitrary (and for parts of Skyloft checks in the same region aren't even grouped together). By default, the model only seems to support alphabetical sorting, but I imagine we could probably write a custom sort order, or even just recreate the model every time there's a change to preserve the order rather than sorting it, however both of those will likely result in a (likely minor) performance hit

lepelog: The change to the check order was an accident that should probably be fixed. Can you just enumerate all the check names and sort on the number? It doesn't need to be displayed

lepelog (approved):

PR #288, opened 2022-12-20
n hints per stone

CovenEsme:
Allows hint distributions to specify how many hints should appear on each Gossip Stone. This ranges from 1-8 hints per stone.

Fewer than 1 doesn't make any sense since, if no hints are desired, one can simply use the 'Junk' hint distribution.

More than 8 can result in hints getting cut off from the end of the list when viewed in-game. I suspect that this is due to a character limit but I haven't confirmed this. 8 hints per stone is likely excessive but does provide room to experiment should anyone wish to do so (e.g. having fewer but more valuable hint stones).

### DIFF
--- a/hints/hint_distribution.py
+++ b/hints/hint_distribution.py
@@ -486,7 +486,7 @@ class HintDistribution:
         return BarrenGossipStoneHint(area)
 
     def _create_junk_hint(self):
-        return EmptyGossipStoneHint(self.junk_hints.pop())
+        return EmptyGossipStoneHint(self.rng.choice(self.junk_hints)
 
     def get_junk_text(self):
         return self.junk_hints.pop()
--- a/hints/hint_distribution.py
+++ b/hints/hint_distribution.py
@@ -121,9 +121,15 @@ class HintDistribution:
     def _read_from_json(self, jsn):
         self.hints_per_stone = jsn["hints_per_stone"]
         # Limit number of hints per stone as there appears to be ~600 character limit to the hintstone text.
-        if self.hints_per_stone <= 0 or self.hints_per_stone >= 9:
+        if self.hints_per_stone >= 9:
             raise ValueError(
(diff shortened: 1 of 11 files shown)


logic/constants.py
```diff
@@ -8,7 +11,18 @@
...
+
+hint_distros = [f for f in listdir(HINT_DISTROS_DIRECTORY) if f.endswith("json")]
+max_hints_per_stone = 0
+for distro_filename in hint_distros:
+    with open(HINT_DISTROS_DIRECTORY / distro_filename) as f:
+        hints_per_stone = json.load(f)["hints_per_stone"]
+        if hints_per_stone > max_hints_per_stone:
+            max_hints_per_stone = hints_per_stone
+
+MAX_HINTS = max_hints_per_stone * NUMBER_OF_HINT_STONES
```
lepelog, line 25:
```suggestion
MAX_HINTS_PER_STONE = 8
MAX_HINTS = MAX_HINTS_PER_STONE * NUMBER_OF_HINT_STONES
```

That also means we can avoid reading all the distros here

hints/hint_distribution.py
```diff
@@ -152,7 +163,7 @@ def start(
 
         self.banned_stones = list(map(areas.short_to_full, self.banned_stones))
         self.max_hints_per_stone = {
-            stone: 0 if stone in self.banned_stones else MAX_HINTS_PER_STONE
+            stone: 0 if stone in self.banned_stones else self.hints_per_stone
             for stone in self.areas.gossip_stones
         }
```
lepelog, line 168: I think we should be consistent with `max_hints_per_stone` and `hints_per_stone`

I would go with `hints_per_stone` cause the number is fixed, stones can't have fewer than that

lepelog (changes requested):

CovenEsme:
The suggested changes have been made :p

lepelog (approved):

PR #292, opened 2022-12-21
New GUI Part 1 (Removing Banned Types and Rearranging/Regrouping Options)

cjs8487:
## Changes
- Removed the Progress Items Tab
  - Options for progress locations that are choices (dropdowns) remain as options. This includes Batreaux Limit, Rupeesanity, and Beedle's Shop
  - Binary progress location choices now must be done via Exclusions
  - **These changes carry no functional changes for what can and cannot be changed about the randomizer**
- Renamed "Additional Settings" tab to "Randomization Settings"
  - Regrouped many settings
    - New "Shuffles" group for options that control shuffling additional parts of the game (shops, rupees, etc.)
    - New "Completion Conditions" group which combines the old "Gate of Time and Horde Door," "Triforce," and "Endgame Bosses" groups.
  - Migrated the "Additional Options" group to a new tab
- New "Additional Options" tab which houses several miscellaneous options, including optional tweaks to the vanilla gameplay, item pool modifications, and additional settings for the Silent Realms
- Moved several other options around
  - The option for Hero Mode is now on the "Logic Settings" tab
  - The options for Chest Dowsing are now on the "Hints" tab
- Renamed permalink to Settings String
- Expanded the width of the option description pane
- Minor change to the UI style

## Technical Changes
- Every component is now named to make skimming the XML and component tree easier. This should also help make resolving merge conflicts in the .ui file easier
- Removed several extraneous containers
- Removed remnants of banned types from the code

## Further Changes
- To maintain the status quo of easy bans for commonly excluded locations, a revamped system for location exclusion is planned, but is beyond the scope of this PR
- Additional changes to strings are needed, but are beyond the scope of this PR
- Several planned/semi-implemented options do not yet have homes in this iteration. That responsibility will fall to the PRs that add them. Almost all currently planned options should have clear homes once this iteration is complete. There is additional space in the Additional Settings tab should it be needed in the future
- (Technical) The main UI control file has grown quite large, and will need to be refactored in the near future to increase maintainability and reduce need for boilerplate code in several circumstances
- The UI needs to be rethemed, and in the process we can give users some amount of customization over their appearance, even if it's as simple as matching their system light/dark mode settings


lepelog (changes requested):
I think it makes sense to update the presets, so that they act the same as previously (now they don't exclude locations at all, which is wrong)

lepelog (approved):

PR #294, opened 2022-12-24
Start item cleanup/start dungeon items

lepelog:
still needs ui


gamepatches.py
```diff
@@ -1552,40 +1552,67 @@ def add_startitem_patches(self):
...
-                    startflags.append(flag)
-            else:
-                startflags.append(flags[item])
-        return startflags
+        start_item_counts = Counter(
+            map(strip_item_number, self.placement_file.starting_items)
+        )
+        # health is calculated in quarter hearts
+        starting_health = 6 * 4
+        starting_health += start_item_counts.pop(HEART_CONTAINER, 0)
```
yannl35133, line 1560: Shouldn't this be multiplied by 4 if the unit is quarters ?

lepelog: good catch

CovenEsme:
Still need to give this a test but would it make sense to have the maps as starting items by default? Would be good for newer players to have the dungeon maps already to help with finding checks.

lepelog:
> Still need to give this a test but would it make sense to have the maps as starting items by default? Would be good for newer players to have the dungeon maps already to help with finding checks.

It might make sense to add them to the startitems for the beginner presets

PR #295, opened 2022-12-24
Boko base

CovenEsme:
Adds access to and from Boko Base! The Shiekah Stone usually found in Boko Base has been added to Eldin Volcano (near the Volcano East bird statue) and offers access to Boko Base. There is another Shiekah Stone in the Boko Base prison that allows players to return to Eldin Volcano. There were plans to have Plats offer this return to Eldin but this would force players to get the `Plats' Gift` check which is less obvious than the Shiekah Stone.

Adds 10 checks: `Plats' Gift` (Mogma Mitts), the 5 b-wheel item chests, the Volcano Summit small chest, `Raised Chest in Volcano Summit` (True Master Sword), `Chest in Volcano Summit Alcove` (Pouch), and the Fire Dragon's SotH Piece.

In addition, the 2 chests that were added to Volcano Summit from Boko Base have been removed now that they can exist in their vanilla locations instead.

Players also start with the map completed to show the b-wheel items.

Currently, players keep their items when entering and leaving Boko Base. There are plans to add an option to have items become restricted when entering Boko Base but this is significantly more complex for minimal benefit.


CovenEsme:
Barring any further bugs being found, Boko Base is fully functional. What follows are my observations regarding a `restrict-boko-base` option where items are restricted whilst on the boko base stage.

Items can be restricted on entry and unrestricted on exit very easily. This is trivial and functions very well.

In the vanilla game, items are unrestricted when obtaining their relevant check. This is achieved with a switch on the item get event that checks for each item's restricted story flag. As a result, placing a different item on these checks doesn't unrestrict the original item.

As far as I can see, there are 3 possible solutions and 2 reasonable solutions.

1) A patches solution would be to utilize the fact that each of these checks sets a scene flag. By placing an AREA around each check, it could be possible to add a Plats object near each check so he pops up and "gives" an unlock for a restricted item. This would involve quite a lot of patches but would be manageable.

2) Create an additional chest type or make chests optionally trigger a generic event that can be patched in `eventpatches.yaml`. This would require asm and would likely be very complicated. It would, however, be the cleanest option - provided it exposes a visual indication that a restriction has been lifted.

3) Patch *every* item get text with a switch that can be modified to unrestrict a restricted item. This is ridiculous and would be utterly unmaintainable.

Dismissing the 3rd option, it would then be possible to randomize the lifting of the restrictions in a similar fashion to how the required dungeons are patched.

With the skills and knowledge of the game that I currently have, I believe I could implement the first solution if this were desired. However, I suspect that those who would like an option to have items restricted when entering Boko Base would be unsatisfied with this solution.

If anyone has any alternative suggestions about how it may be possible to add a `restrict-boko-base` option, I would be very willing to helping to help implement it.

CovenEsme:

PR #296, opened 2022-12-24
Removing Hero Mode in Favor of Split Options.

elijah-thomas774:
Hero Mode is now Removed!
Dont worry, its benefits AND drawbacks are now split for people to have an option!!
Now you can enjoy  the upgraded skyward strike and still have heart pickups available from the beginning.

Right now it is:
Damage Multiplier (1-255) 
Heart Pickups (like normal vs Hero mode)
Air meter depletion normal or fast
Upgraded Skyward Strike

The options need testing to see if they work properly as only quick testing was made as so the game launched and did not crash.


CovenEsme:

PR #297, opened 2022-12-26
Fix random starting item not same for all players

CovenEsme:
Fixes a bug where the players using the same seed that has the `random-starting-item` option enabled may not get the same item. This is due to a use of set comprehension making the random choice result in a potentially different starting item for the same seed.

### DIFF
--- a/logic/randomize.py
+++ b/logic/randomize.py
@@ -200,9 +200,11 @@ class Rando:
 
         if self.options["random-starting-item"]:
 
-            possible_random_starting_items = RANDOM_STARTING_ITEMS
-            for starting_item in self.options["starting-items"]:
-                possible_random_starting_items.remove(starting_item)
+            possible_random_starting_items = [
+                item
+                for item in RANDOM_STARTING_ITEMS
+                if item not in self.options["starting-items"]
+            ]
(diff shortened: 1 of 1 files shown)


logic/randomize.py
```diff
@@ -199,9 +199,10 @@ def randomize_starting_items(self):
                 starting_items.add(item)
 
         if self.options["random-starting-item"]:
-            possible_random_starting_items = tuple(
-                set(RANDOM_STARTING_ITEMS) - set(self.options["starting-items"])
-            )
+
+            possible_random_starting_items = RANDOM_STARTING_ITEMS
+            for starting_item in self.options["starting-items"]:
+                possible_random_starting_items.remove(starting_item)
```
lepelog, line 205:
```suggestion
            possible_random_starting_items = [
                item
                for item in RANDOM_STARTING_ITEMS
                if item not in self.options["starting-items"]
            ]
```

Otherwise it produces an error when you start with multiple beetles for example, also don't modify the global variable, this will break once you roll multiple seeds

lepelog (changes requested):

lepelog (approved):

PR #298, opened 2022-12-27
Remove camera panning during Fledge's Gift

CovenEsme:
Adds a text sequence onto Fledge's Gift to display the required dungeons for the seed. Additionally displays the same required dungeon text when talking to Fledge again in the Academy corridor.

Removes several panning camera changes during Fledge's Gift.


CovenEsme:
Fledge no longer tells you your required dungeons. The camera pans have been cut out as well as the text telling you to look at the notice board (since that's being moved to Fi). Fledge is also closer to Link's room so he doesn't look like he's yelling across the hallway xD

PR #299, opened 2022-12-28
Remove mogma cutscene before circles

Muzugalium:


CovenEsme:
Works great 👀

lepelog:
Iirc this was previously not removed because that's the flag that opens batreaux door on skyloft. Idk if that is still relevant

CovenEsme:
Personally, I don't really think it's relevant. We could make the mogma be popped out of the ground and make it so talking to him gives the flag tho? Would this be desirable?

Muzugalium:
I dunno if it's relevant since that flag is a start flag anyway but I've made it so the mogma stays up and gives you that flag if you manually speak to him ^^

PR #306, opened 2022-12-31
Fi menu overhaul

Muzugalium:


Muzugalium:
Found a bug with the dungeon status where maps are always shown as unobtained, will try and fix tomorrow

asm/custom-functions/src/lib.rs
```diff
@@ -219,6 +238,77 @@ fn handle_bk_map_dungeonflag(item: c_ushort) {
...
+fn rando_text_command_handler(_event_flow_mgr: *mut c_void, p_flow_element: *const FlowElement) {
+    let flow_element = unsafe { &*p_flow_element };
+    match flow_element.param3 {
+        71 => {
+            let dungeon_index = flow_element.param1;
+            let completion_storyflag = flow_element.param2;
+            let mut key_count = dungeon_global_key_count(dungeon_index);
+            if dungeon_index == 14 {
+                unsafe { key_count = getKeyPieceCount();}
+            }
```
lepelog, line 257:
```suggestion
            let key_count = if dungeon_index == 14 /* ET */ {
                unsafe { getKeyPieceCount() }
            } else {
                dungeon_global_key_count(dungeon_index)
            };
```

gamepatches.py
```diff
@@ -1648,42 +1651,110 @@ def add_required_dungeon_patches(self):
...
             combined = ""
 
             for part in required_dungeons_text.split(" "):
-                if len(cur_line + part) > 27:  # limit of one line
+                # limit of one line (notice board = 27, textbox = 36)
+                if len(cur_line + part) > 36:
                     combined += cur_line + "\n"
                     cur_line = part + " "
                 else:
                     cur_line += part + " "
```
lepelog, line 1681: Use `break_lines` from `textbox_utils`(it didn't exist when this code was written, but would be a good cleanup)

lepelog (changes requested):

gamepatches.py
Muzugalium: Done ^^

asm/custom-functions/src/lib.rs
Muzugalium: Also done ^^

CovenEsme:

PR #309, opened 2023-01-01
FS back half lava flow startflag

Muzugalium:


lepelog:
I don't think this should be the default, it needs an option at the very least imo

CovenEsme:
Looks good to me. Tested in a co-op seed and worked as intended :p

PR #311, opened 2023-01-06
Open Lake Floria

CovenEsme:
Adds new options for controlling access into Lake Floria. These are:
* `Floria Gates` (requires Yerbal and drawing on the floria door)
* `Talk to Yerbal` (requires talking to Yerbal only)
* `Vanilla` (logically requires talking to Yerbal and drawing on the floria door)
* `Open` (floria door is open from the start of the game)

The `Floria Gates` and `Talk to Yerbal` options disable the loading zones into Lake Floria until their requirements are met. This means that the Early Lake Floria tricks will not work.


CovenEsme:
Both "Vanilla" and "Talk to Yerbal" options have been tested with no issues noted.

patches.yaml
```diff
@@ -1727,6 +1734,65 @@ F100: # Faron main area
...
+    objtype: STAG
+  - name: Patch Main floria entrance
+    onlyif: Option "open-lake-floria" Is "Floria Gates"
+    type: objpatch
+    id: 0xFDB6
+    layer: 0
+    room: 0
+    objtype: STAG
+    object:
+      trigscenefid: 50
```
lepelog, line 1766: sceneflags are dangerous because of RBM, is there a storyflag connected to the gate opening that can be used here

lepelog (changes requested):

patches.yaml
CovenEsme: Unfortunately, no. The door checks for sceneflag 49 (cs of the doors opening) and sets sceneflag 50 (doors open). We could edit the flags to be ones that can't be RBM'd. There are quite a few that are free.

Otherwise, we'd could find the scene flag check in the DoorDun actor and add an extra check for just the Lake Floria gates. That might be complicated tho since the DoorDun actor isn't unique to this situation.

CovenEsme:
Open now only opens the Lake Floria door after obtaining scale. If scale is found in main Faron, a reload is required for the door to open. I tried to get it to update immediately but this seems to be more complicated than the light pillars (which have previously been patched for a similar issue).

CovenEsme:
Tested the "Open" setting in co-op seed. Bug found where obtaining the Water Scale didn't correctly open the Lake Floria door. Bug has been fixed (incorrect scene flag index used).

CovenEsme:
The Scale behaviour isn't consistent for some reason. I've reverted the open setting to have the Floria gates open from the start. I've also patched the main Floria loadzone to only be active after getting Water Scale (open setting only). This acts as softlock prevention if you run off the floria entrance platform without scale. ELF still works with the alternative loadzone when you don't have scale.

PR #312, opened 2023-01-06
commit flagmanagers when setting startflags

lepelog:


CovenEsme:

PR #316, opened 2023-01-08
Shorten Golo text in Lanayru caves

Muzugalium:


CovenEsme:

PR #317, opened 2023-01-08
fix banning checks when there are dependencies

lepelog:


yannl35133:
Manually editing the requirements out of logic.py isn't a clean way to fix this

CovenEsme:
Tested in race today - no issues found. This is a pressing issue that's been causing problems in races. Clean fix or not, a fix is needed and this works.

PR #321, opened 2023-01-10
Sword Pedestal in Goddess Statue

CovenEsme:
Adds back the Sword Pedestal in the Goddess Statue and replaces the 2 small chests. Pulling out the Goddess Sword now gives 2 checks: one for the vanilla Goddess Sword and another for the Emerald Tablet.

There was a weird inconsistent bug where sometimes Fi wouldn't give certain items. I'm 95% sure I've squashed this irksome bug but it's difficult to be sure with this sort of thing.


CovenEsme:

PR #325, opened 2023-01-15
🌟 Add Starry Skies cosmetic option 🌟

CovenEsme:
Adds stars to always be drawn onto the background when the sky is visible. This works on the surface, the Sky, and Skyloft. Stars are added both during the day and the night :D (yep, they appear on the title screen)


CovenEsme:
Has been tested in several races. Works as intended.

Having the star count too high can cause areas to lag. I've added a warning in the tool tip about this. For the default star count, there doesn't appear to be any(?) lag on dolphin. For console, the problem areas are Boko Base and BiT Skyloft.

CovenEsme:
Tested in another co-op seed - works fine.

options.yaml
```diff
@@ -464,6 +464,24 @@
...
+  type: boolean
+  default: false
+  permalink: false
+  cosmetic: true
+  help: Makes stars appear in the sky - even during the daytime.
+  ui: option_starry_skies
+- name: Number of stars in the sky
+  command: star-count
+  type: int
+  default: 700
```
lepelog, line 478: I think there should be min and max here

PR #327, opened 2023-01-16
always patch layers that can be modified

lepelog:
this fixes issues when there are big models added from previous runs, since they weren't removed previously

Kinda jank


CovenEsme:
Tested in a co-op seed - works fine.

PR #329, opened 2023-01-18
fix broken banned locations, rename slingshot

YourAverageLink:
Essentially reverts #317 in favor of removing the inlined requirement for Dodoh. Slingshot seemed to be unrelated, and more related to its name, so it was renamed to Kikwi Elder's Reward. Should also fix nonprogress item calculation and the excluded locations sometimes having progress bug.


CovenEsme:
Tested in a co-op seed - works fine.

PR #330, opened 2023-01-19
Delay computation of SotS items

yannl35133:


CovenEsme:
This still needs the progress bar text changing, right? I didn't notice any difference when testing it.

yannl35133:
It should work, but SotS items are still always computed for the spoiler log, so it's currently useless

PR #331, opened 2023-01-19
Fix bug with Fire Dragon's Lair loadzone

CovenEsme:
Fixes an issue where re-entering the Fire Dragon's Lair from the post-boko base VS layer would put Link on F211 layer 0 without a loadzone to exit again. Resolves the issue that Fireworkspinner encountered in a co-op race on 2023-01-18.


CovenEsme:
Tested in a co-op seed - worked fine.

PR #340, opened 2023-01-24
Logic Tab Improvements - Tricks Dialog and Free Text Search

cjs8487:
Improves the Logic tab of the randomizer. Makes the following changes
- Tricks are migrated to a popup window and removed from the rest of the tab
- Include/Exclude locations have been expanded to fill the vacated space
- Adds a free text search to both the included and excluded locations lists
  - Text searches can be done independently on both views and do not impact each other
  - Moving an item from to a destination view where the search would hide it is supported (item will be hidden by default)
  - Searches are case insensitive, and will match any check name containing the string. It is not limited to word boundaries or any other related limitation

## Technical Notes
- The serialization process for **all** ListViews has been altered
- Sorting is correctly applied to all ListViews whenever they have an item added to it

## Future Notes
- Dropdown/text->category filtering is still planned for the future, but did not make the cut for this PR in the interest of getting this important first usability improvement rolled out
- Include/Exclude All is still planned, but remains unimplemented here

### DIFF
--- a/gui/randogui.py
+++ b/gui/randogui.py
@@ -475,9 +475,9 @@ class RandoGUI(QMainWindow):
             self.options.set_option("enabled-tricks-bitless", [])
             self.options.set_option("enabled-tricks-glitched", [])
 
-        # self.options.set_option(
-        #     "excluded-locations", self.get_option_value("excluded_locations")
-        # )
+        self.options.set_option(
+            "excluded-locations", self.get_option_value("excluded_locations")
+        )
 
         self.save_settings()
         self.ui.permalink.setText(self.options.get_permalink())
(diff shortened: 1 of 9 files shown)


cjs8487:
Now Fixed:
- Fixed undefined behavior in specific cases within the sorting algorithm
- Fixed incorrect behavior when using the "Cancel" button on the Tricks Dialog

Added:
- All list based selectors now automatically selected the next valid element in the list after movement. This allows quick movement of multiple successive elements in the list

CovenEsme:
Tested originally and just tested the fixes now. Cancel and Okay work properly (including just closing the window). Seems good to me :p

PR #343, opened 2023-02-08
Add option to patch BiT

CovenEsme:
Adds an option to force the RELOADER_TYPE when loading the title screen so that it is no longer possible to perform the Back in Time (BiT) glitch. Additionally, removes the now useless "Hero Mode" ui toggle.


options.yaml
```diff
@@ -434,6 +434,12 @@
     Basic tells if the reward item is any progress item. Advanced tells if the trial is WotH, Barren, or potentially required.
     Direct tells the item on the trial directly.
   ui: option_song_hints
+- name: Patch BiT
+  command: patch-bit
+  type: boolean
+  default: false
+  help: If enabled, forces the title screen to reload fully - making the Back in Time tricks impossible.
+  ui: option_patch_bit
```
lepelog, line 442: Would it make sense to merge this with `fix-bit-crashes` into a single setting? With choices like
- no changes
- make BiT not crash
- make BiT not possible at all
but better names obviously

CovenEsme: Could do something like:
* Disable BiT
* Vanilla
* Fix BiT Crashes

options.yaml
```diff
@@ -434,12 +434,20 @@
...
+  command: bit-patches
+  type: singlechoice
+  bits: 2
+  choices:
+    - Disable BiT
+    - Vanilla
+    - Fix BiT Crashes
+  default: Vanilla
+  help: "Changes how the Back in Time (BiT) glitch works.
+        Disable BiT: Forces the title screen to reload fully - making all BiT tricks impossible.
```
lepelog, line 447:
```suggestion
        Disable BiT: Makes it impossible to activate BiT.
```
`Forces the title screen to reload fully` is not accurate at all, the title screen loads fully and properly in any case, the difference is really if link respects the params of the entrance, that are supposed to trigger an infinite cutscene

options.yaml
```diff
@@ -434,12 +434,20 @@
...
+        Vanilla: Keeps the vanilla game behaviour where BiT is possible but Eldin BiT (and others) will still crash.
+        Fix BiT Crashes: Does not load some arcs in BiT when all file slots are filled to make all areas accessible in BiT."
```
lepelog, line 449: Not sure how to word it, but maybe it should stress that all the file slots need to be filled *before* activating BiT?

lepelog (changes requested):

CovenEsme:
Has been tested in several seeds and races. Patched BiT worked in the monthly no-BiT race and Vanilla worked in subsequent BiT races.

PR #346, opened 2023-02-08
Update MAP_CHECKS to use correct check names

CovenEsme:
Updates the `MAP_CHECKS` constant to use the correct check names for the vanilla map chests. This allows the `map-mode` to be able to be set to "Vanilla" again without erroring.


CovenEsme:
Fixed issue where starting with some but not all maps (or boss keys maybe? it works now but I didn't check if it was broken before my changes 🤷‍♀️) and having `map-mode` set to vanilla would throw an error. Now, any maps or boss keys you start with are added as starting items and the remaining maps and boss keys are placed in their vanilla locations.

Additionally, I've added more info to the 'ValueError's in `logic.py` so it's easier to debug issues with them.

logic/logic.py
```diff
@@ -66,33 +66,72 @@ def copy(self):
...
+        other.items = {
+            item: location
+            for item, location in other.items.items()
+            if item not in start_items
+        }
+        other.locations = {
+            location: item
+            for location, item in other.locations.items()
+            if item not in start_items
+        }
```
yannl35133, line 116: This is not a good way to fix this because :
- Placements are immutable, yet you're changing fields
- Start items have a location, called "Start Item" (IIRC)
- This is not the right place to have such logic, this belongs in `randomize.py`

CovenEsme: * Where are `Placements` defined to be immutable? afaik, python doesn't have a robust concept of immutability (which this doesn't use).
* This is true and is what I use to determine if an exception is needed.
* Fixing this in `randomize.py` would be more complex and require duplicated code for all instances where items can be placed in their vanilla locations and as starting items. This fixes the issue in one location in a way that is future-proof.

IMO, the logic should """just handle""" placing a start item in its vanilla location. It's easier to maintain and means that anyone adding options for vanilla item placements doesn't have to deal with this issue.

yannl35133: It's not defined as immutable, it's handled as immutable. The code makes and relies on this assumption.
Also, you can put that logic in `randomize.py` right before returning from the initialisation; everything about start items and initial placements (vanilla locations) will already have been handled.

CovenEsme: The error from the conflict between the start items and the vanilla items happens during the init. I'm not sure where you would want the logic to go? afaik, maps and boss keys would have to be handled separately. Where exactly are you suggesting the logic should be moved to?

yannl35133: Oh right, it raises an exception, I forgot that part. Then you could make it a dedicated function, but this doesn't belong in `__or__`

CovenEsme: ngl, this doesn't seem like an issue to me. If you really care then feel free to leave a code suggestion or create a PR onto this branch to better show what you mean. I don't see why defining a case that takes priority when using the or operator doesn't belong in `__or__`.

A function would need to be called both for maps and boss keys which is manageable for now but means that anyone wanting to add other vanilla options (like vanilla single gratitude crystals or vanilla dungeon items) would also have to know about and remember this quite obscure bug.

yannl35133: This is not a bug, it's semi-obscure but totally intended. Vanilla means pre-placed, which is supposed to be incompatible with placing an item in the starting inventory

CovenEsme: It's undesired behaviour which makes it a bug. There are 3 possible ways to handle this: 1) make starting items take priority over vanilla placement 2) make vanilla placement take priority over starting items 3) display and error saying that this combination of settings is not possible. The current code in master does none of these.

yannl35133: I meant that it is not a bug of `__or__`. As handled currently, the options would be incompatible, and the issue is that it is not reported to the user.

CovenEsme: `__or__` is bugged due to its lack of meaningful error messages. The current handling of conflicting starting items and vanilla placements in `__or__` has been discussed (https://discord.com/channels/767090759773323264/767092977755488267/1072999758974689290) and a different behaviour is desired. Whether you view this as a bug or not is debatable (and not one that matters). Regardless, these bugs/issues originate in `__or__` and fixing them there makes the most sense.

You seem to have an alternative solution in mind but it seems to have the limitation that makes it less future-proof than the solution I've implemented here. Again, if you can implement something better, please do. However, it seems that you just don't like how I've done it.

yannl35133: `__or__` is a library function, it is not responsible for generating errors to the user.
That different behaviour was desired by you, your presentation looks slightly disingenuous; this is not the current behaviour, and it is expected that rather consequent changes be needed to achieve this new behaviour.

To implement your desired behaviour, my solution would be not to preplace the vanilla items and instead add the restrictions that the items can only be placed at their vanilla location (using the restriction mechanism instead).

And yes, I "just don't like it", because breaking established invariants is the easiest way for a bug to sneak in (see lepe's PR for banned locations)

CovenEsme: I said lack of meaningful error messages, not error messages to the user. `raise ValueError` with no error message makes debugging much harder. There's no indication why the error is raised. `__or__` doesn't raise a meaningful error which is bad for devs. The fact that this isn't caught somewhere else and a user facing error isn't thrown is bad for users. Both are decisions you made without discussion. The current behaviour is bugged.

The existing behaviour was decided by yourself. I fixed an issue where the map checks were incorrectly named and Lepe pointed out that it still errors if you start with some but not all maps. The discussion was about if this is desired. Yes, I suggested that it wasn't but Lepe agreed. I haven't just decided this on a whim.

Not preplacing vanilla items would work but only insofar as it avoids the bugs with preplacing items. I understand the appeal of not breaking the invariants but it's already bugged.

CovenEsme: Also, this doesn't just affect maps and boss keys. Vanilla Beedle also has issues when starting with a pouch or the bug net. My solution fixes this without it even being known about. It's more generalizable than changing each use of `norm_vanilla`.

cjs8487: > This is not a bug, it's semi-obscure but totally intended. Vanilla means pre-placed, which is supposed to be incompatible with placing an item in the starting inventory

Following up on this - Vanilla does *not* mean preplaced, it means the randomization location is predetermined. Preplacement is the mechanism by which predetermined locations are resolved. You’re once again trying to push your personal headcanon upon people who have largely agreed that it is wrong, and just because it may function this currently doesn’t mean it must remain this way or that it even was the correct implementation in the first place. Behavior that causes crashes because of completely reasonable interactions is a bug, regardless of if it was intended behavior or not. Failures are almost never an acceptable “intentional behavior.”

Beyond that, having the two be incompatible just doesn’t make sense, let alone being objectively wrong. All vanilla placement options mean that the items that are in the pool are vanilla, and that has always been the case

yannl35133: Lacking error descriptions cannot be a bug, it is just annoying. Good on you for fixing that still. This was no "decision" of mine, I just never needed them.

The existing behaviour was up to interpretation at the time of old logic, I cemented it one way here. Apparently I was the only one who didn't know that the way I cemented it was not the "correct one", which is a shame.

Please explain to what bugs there currently are in that function. That preplaced items and starting items are incompatible is not a bug, by the definition of "preplaced".

I never opposed to changing (restoring ?) the interpretation of "vanilla", I just oppose doing it in a way which breaks invariants. I proposed a correct way to do

lepelog: The immutability can be fixed by just using a temporary value. I agree that the inputs of `__or__` should be immutable, since that's the behavior I would expect from just using `a | b`.

Placement being immutable is never documented anywhere, if stuff breaks if it's mutated **that should be documented somewhere**. We can't read each others minds.

If this is not the correct way to handle this, @yannl35133 feel free to suggest a better way (like an actual implementation, not something vague like "do it there"), that way we can have a meaningful discussion about it, cause I'm more concerned with this behaving as intended and not with implementation details.

Side note, is it possible to just restrict the item to a single location, but not actually placing it, could that work? Cause then it would be fine if the item is part of startitems

yannl35133: If any "vanilla" option should restrict to a single location instead of forcing it, then it would look like :
```diff
- def norm_vanilla(list: List[str]):
+ def norm_vanilla(locations: List[str]):
    def norm_keys(norm: Callable[[str], EIN], checks: Dict[EIN, Any]) -> Placement:
-         list2 = map(norm, list)
-         dict = {k: EIN(checks[k]["original item"]) for k in list2}
-         return Placement(locations=dict, items={v: k for k, v in dict.items()})
+         locs2 = map(norm, locations)
+         restriction = {EIN(checks[k]["original item"]) for loc in locs2}
+         return Placement(item_placement_limit=restriction)

    return norm_keys
```

CovenEsme:
Assuming everyone's okay with the new fix, this should be good.

logic/placements.py
```diff
@@ -31,7 +33,18 @@ def norm_keys(norm: Callable[[str], EIN], checks: Dict[EIN, Any]) -> Placement:
     return norm_keys
 
 
-SINGLE_CRYSTAL_PLACEMENT = norm_vanilla(SINGLE_CRYSTAL_CHECKS)
+# Restricts an item to only be placeable in its vanilla location.
+# Does not produce an error if this is not possible (e.g. if the item is a starting item).
```
yannl35133, line 37:
```suggestion
# This is always possible
```

yannl35133 (approved):
Small nitpick, but I'm okay with this

logic/placements.py
```diff
@@ -169,5 +182,5 @@ def norm_values(norm: Callable[[str], EIN]):
 )
 
 
-VANILLA_RUPEES = norm_vanilla(RUPEE_CHECKS)
-VANILLA_QUICK_BEETLE_RUPEES = norm_vanilla(QUICK_BEETLE_CHECKS)
+VANILLA_RUPEES = norm_restrict_vanilla(RUPEE_CHECKS)
+VANILLA_QUICK_BEETLE_RUPEES = norm_restrict_vanilla(QUICK_BEETLE_CHECKS)
```
lepelog, line 186:
```suggestion
VANILLA_RUPEES = norm_force_vanilla(RUPEE_CHECKS)
VANILLA_QUICK_BEETLE_RUPEES = norm_force_vanilla(QUICK_BEETLE_CHECKS)
```
Same as for beedle's shop: Rupees are not patched at all when rupeesanity is set to vanilla

logic/placements.py
```diff
@@ -31,7 +33,18 @@ def norm_keys(norm: Callable[[str], EIN], checks: Dict[EIN, Any]) -> Placement:
...
+def norm_restrict_vanilla(locations: List[str]):
+    def norm_keys(norm: Callable[[str], EIN], checks: Dict[EIN, Any]) -> Placement:
+        locs2 = map(norm, locations)
+        restriction = {EIN(checks[loc]["original item"]): loc for loc in locs2}
+        return Placement(item_placement_limit=restriction)
+
+    return norm_keys
+
+
+SINGLE_CRYSTAL_PLACEMENT = norm_restrict_vanilla(SINGLE_CRYSTAL_CHECKS)
```
lepelog, line 47:
```suggestion
SINGLE_CRYSTAL_PLACEMENT = norm_force_vanilla(SINGLE_CRYSTAL_CHECKS)
```

logic/placements.py
```diff
@@ -64,23 +77,23 @@ def norm_values(norm: Callable[[str], EIN]):
     "Beedle's Shop - 50 Rupee Item",
     "Beedle's Shop - 1000 Rupee Item",
 ]
-VANILLA_BEEDLE_PLACEMENT = norm_vanilla(BEEDLE_CHECKS)
+VANILLA_BEEDLE_PLACEMENT = norm_restrict_vanilla(BEEDLE_CHECKS)
```
lepelog, line 80:
```suggestion
VANILLA_BEEDLE_PLACEMENT = norm_force_vanilla(BEEDLE_CHECKS)
```
This can be problematic: Currently, beedle isn't patched at all, when vanilla placement is selected (and it also makes pouch necessary to buy the pouch upgrades). Even if a different item than the vanilla one is randomized in beedle's shop, it wouldn't appear in game

lepelog (changes requested):
The changes for dungeon items look good, also :+1: for improving error messages

logic/placements.py
yannl35133: Now the comment saying `norm_force_vanilla` is unused can be removed

CovenEsme:
Requested changed made :p

PR #347, opened 2023-02-10
Decouple skeep checks from first chest (yeet bars)

CovenEsme:
Removes the bars in the first room of Sky Keep preventing access to the control panel. Also removes the event that plays showing off the bars opening (the sceneflag is still set as a result of opening the chest). Means that the first chest isn't required for the rest of the other Skeep checks.


Muzugalium:
Tested getting all checks (3 triforce pieces, small key chest, and rupee in FS room) before getting first chest, works

PR #348, opened 2023-02-10
Remove bipping after Owlan's Gift check

CovenEsme:
Removes the bipping sound after getting the Owlan's Gift check.


CovenEsme:
Has been tested in races and hasn't caused any problems.

PR #349, opened 2023-02-10
DER Fixes

CovenEsme:
Skips over the SSH crest cutscene (missed in the previous cs yeeting spree).


CovenEsme:
The DER fixes (inc. sadship crest cs fix) have been tested and work. Just needs the Sky Keep bug testing as well as the ET door cs skip.

Muzugalium:
Tested skykeep dive, works

lepelog (approved):

PR #351, opened 2023-02-15
Expand arc replacements to work with all arcs

Muzugalium:
-The remaining unpatched arcs are now supported by arc_replacements
-Due to some duplicate arc names, there are specific names filenames required for some of the arcs to get around this- fortunately they're for arcs that aren't commonly modified anyway (motion plus tutorial cursor arc and the Spanish and French stage text arcs), see changelog for specifics
-The arc replacements folder now supports sub folders so people can organise their arcs freely
-The existing patches to DoButton.arc and Title2D.arc that handle the custom textures for the title screen and dowsing icons now pull from modified_extract instead of actual_extract so they don't fully overwrite replaced arcs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,16 @@
   - The 2 small chests previously found within the Goddess Statue have been removed
 - Removed all skippable cutscenes except boss intro cutscenes (by CovenEsme)
   - When starting a new file, Link will now spawn directly in his room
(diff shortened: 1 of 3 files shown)


CovenEsme:
Works as intended. Tested with ObjectPack arcs, commonly replaced arcs (DoButton, FileSelect, MenuPause), misc arcs (GoddessCube, SkyCmn), stage arcs (GetHeartUtuwa, PutHeartUtuwa).

PR #353, opened 2023-02-17
New Multichoice Implementations for Small Lists and Conditionals

cjs8487:
## Feature Details
Adds custom widget implementations for additional multichoice options, specifically options with small choice lists and conditionals for the option that exist outside the multichoice.

## Technical Notes
- This introduces an arbitrary widget replacement paradigm used to replace basic blank widgets placed via the designer with the custom widgets at runtime, this determination is made based on the type details of the option that matched the widgets name
- Adds two option parameters to multichoice options - "trivial" for small lists and "conditional" for conditional multichoice. There are mutually exclusive, and trivial is evaluated first. These must be used in order for the replacement paradigm to take effect and for the option to be utilized by the UI (it will be ignored otherwise, as custom code is required for "complex" multichoice options
- The first 4 choices in the choice list for a conditional multichoice *must* be the expected 4 options defined in the conditional multichoice class. These are ignored when compiling the option list. Consequently, the number of bits required for the option is 3 greater than the maximum size of the final list (more details on this later)

## Design Notes
I haven't taken the design process of this particular PR lightly, and I'm writing down al my design thoughts so there can be meaningful discussion around it while I finish the implementation. This is the first time we have had to deal with "sub-types" of options. I've elected to follow the more traditional parent-child model for this - functionally these new types of multichoice are identical to the current implementation of multichoice options (with a small exception for conditionals which I'll go into detail on later). Following this model allows the current code for writing/reading permalinks to remain unchanged and avoids risks with other places that may depend on specific option types.

### Subtype Descriptions
A trivial multichoice option is any option with an arbitrarily small number of choices. What this number is is well beyond the scope of this discussion, although I'd posit that it's somewhere between 10 and 15.

Conditional multichoice options are effectively two very tightly coupled options - a singlechoice option for the mode and a multichoice option that is only used when a specific option is chosen in the singlechoice. Typically all conditional multichoice options have the same 4 options for the singlechoice - "None," "All," "Random," and "Choose" with Choose being the option that enables the multichoice. Interestingly, 3 of the 4 mode options can be ignored by the randomizer completely - they are essentially metadata, since they have identical representations with the multichoice list alone

### Challenges
This PR posed many challenges - one of the largest of which was the lack of native multichoice style widget in Qt. The lack of native widget means that one either has to be created or improvised, which led to needing to create custom data models and wire everything.

The other largest challenge was designing a solution for conditional multichoice options. The challenge lies in the representation of the mode option, since it is essentially just noise. However, the UI does not have the capacity to properly generate a seeded random list, the responsibility has to fall to the randomizer, and can only be done after clicking the Randomize button, and thus it has to compiled as part of the options. This also solves the problem of restoring UI state from a file or permalink and prevents ambiguous situations such as "Choose' with an empty choice list vs "None" or "Choose" with all choices chosen and "All." As a result, the composite option data for a conditional multichoice will always be a list, where the 0th index is the singlechoice option and the remainder of the list is the multichoice selection list.


cjs8487:
This PR should now be fully functional, however I'm leaving as a draft until the discussion wraps up. The codebase isn't particularly testable as it stands now since there are no options that fulfill the criteria for creation of the new widgets. To test, you can add a blank widget somewhere in the UI with the designer called `option_test` and add one of the following to `options.yaml`

```
- name: Test
  command: test
  type: multichoice
  conditional: true
  default: ['Off']
  choices:
    - 'Off'
    - Random Selection
    - Choose
    - All
    - Test 1
    - Test 2
    - Test 3
  help: Test option
  ui: option_test
```
for a conditional multichoice or
```
- name: Test
  command: test
  type: multichoice
  trivial: true
  default: []
  choices:
    - Test 1
    - Test 2
    - Test 3
  help: Test option
  ui: option_test
```
for a trivial multichoice.

cjs8487:
This should be up to date and ready now

PR #354, opened 2023-02-18
add ability to plrstags to trigger an entrance after death without reverting the inventory and apply that to boko base

lepelog:
This fixes the issue when getting an item from the slingshot chest in boko base after the bomb bag chest would revert that item after death.

It needs to be tested, that this doesn't break PlRsTag elsewhere (like around the AC BK Chest and during the SSH mast sequence)


CovenEsme:
Tested the Boko Base bug as well as the AC BK chest and SSH mast (including mast skip). All works well :p

PR #358, opened 2023-03-01
remove unchangable fill algorithm option

lepelog:


yannl35133:
It was changeable, using the CLI

lepelog:
which nobody seriously uses, but the bigger problem is that this just gets enabled if some bits of a permalink line up badly, and there is no warning or anything

PR #359, opened 2023-03-08
Fi Hints

YourAverageLink:
Adds Fi as a hint source
- She holds gossip stone hints, and hints are placed on her first before filling any remaining stones
- They may be accessed from her menu (Information -> Hints)


logic/hints.py
```diff
@@ -110,22 +110,36 @@ def do_hints(self, useroutput: UserOutput):
...
             check_hint_status,
         )
-        hintstone_hints = self.dist.get_hints()
+        fi_hints, hintstone_hints = self.dist.get_hints()
         self.useroutput.progress_callback("placing hints...")
         hintstone_hints = {
             hintname: hint for hint, hintname in zip(hintstone_hints, HINTS)
         }
+        fi_hints = {hintname: hint for hint, hintname in zip(fi_hints, HINTS)}
+        self.logic.placement.fi_hints |= set([hintname for hintname in fi_hints])
```
yannl35133, line 119: Where is this used ? `HINTS` are the gate to requirements for hint logic, if you make it no-logic you shouldn't bother with those. `placement.fi_hints` looks unused except locally, so i don't understand why it exists a a full field
```
        placed_fi_hints = {
            FI_HINTS: FiHintWrapper(fi_hints)
        }
```
This should do

yannl35133 (changes requested):
Remarks on the logic part

hints/hint_distribution.py
```diff
@@ -119,16 +119,21 @@ def read_from_str(self, s):
...
             )
-        elif self.hints_per_stone <= 0:
+        if (self.fi_hints < 0) or (self.hints_per_stone < 0):
             raise ValueError(
-                "Selected hint distribution must have at least 1 hint per stone. "
+                "Selected hint distribution must not have less than 0 Fi hints or hints per stone."
+            )
+        elif (self.hints_per_stone == 0) and (self.fi_hints == 0):
+            raise ValueError(
+                "Selected hint distribution must have at least 1 hint per stone or at least 1 Fi hint. "
```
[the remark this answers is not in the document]
YourAverageLink: Holdover from when 0 hint stones used to error... I guess it doesn't need to stay.

logic/hints.py
```diff
@@ -110,27 +110,33 @@ def do_hints(self, useroutput: UserOutput):
...
```
lepelog, line 119: This seems unused? Remove

logic/logic.py
```diff
@@ -44,6 +44,7 @@ class Placement:
     locations: Dict[EIN, EIN] = field(default_factory=dict)
     items: Dict[EXTENDED_ITEM_NAME, EXTENDED_ITEM_NAME] = field(default_factory=dict)
     stones: Dict[EIN, List[EIN]] = field(default_factory=lambda: defaultdict(list))
+    fi_hints: Set[Hint | GossipStoneHintWrapper] = field(default_factory=set)
```
lepelog, line 47: Unused?

gamepatches.py
```diff
@@ -1728,6 +1728,10 @@ def add_fi_text_patches(self):
         else:
             required_dungeons_text = break_lines(", ".join(colourful_dungeon_text), 44)
 
+        fi_hint_chunks = []
+        for i in range(0, len(self.placement_file.hints["Fi"]), 8):
```
lepelog, line 1732:
```suggestion
        for i in range(0, len(self.placement_file.hints[FI_HINTS]), 8):
```

hints/hint_distribution.py
```diff
@@ -166,7 +166,7 @@ def start(
             stone: 0 if stone in self.banned_stones else self.hints_per_stone
             for stone in self.areas.gossip_stones
         }
-        self.nb_hints = sum(self.hints_per_stone.values())
+        self.nb_hints = sum(self.hints_per_stone.values()) + self.fi_hints
         assert self.nb_hints <= MAX_HINTS
```
lepelog, line 170: "MAX_HINTS" now also has the max fi hint count, but that means a lot more gossip stone hints are possible now, that isn't intended, right?

hints/hint_types.py
```diff
@@ -249,15 +294,48 @@ def to_spoiler_log_json(self):
...
         return self.text
 
     def to_spoiler_log_text(self, norm) -> str:
         return self.text
 
     def to_spoiler_log_json(self):
         return {"text": self.text, "type": self.hint_type}
+
+
+@dataclass
```
lepelog, line 314: I think this can be removed, cause you override `__init__` anyways

lepelog (changes requested):

YourAverageLink:
Since no complete alternative has been given for how to implement these, I'll mark this as ready for review rather than draft.

CovenEsme:
Has been tested in many seeds with different settings (blitz tourney settings, mine and Muzu's co-op settings, co-op + tadtones, co-op + tadtones - cubes) and seems to hold up well.

I've also genned a few logs with some more extreme settings to see how the edge cases are handled.

Seeds with no locations excluded seem to work. However, the always *and* sometimes hints take precedence over the path hints. Genning a seed with 3 or more dungeons required and no locations excluded only gives 2 path hints. Not sure if this is the intended compromise.

Seeds with no dungeons, no batreaux, beedle cheap, all the always and sometimes hinted locations excluded, all starting items works fine. You get 2 barren hints and the rest are filled with random hints.

Not sure what the concerns with the implementation details were so I can't comment on those.

tl;dr: always and sometimes(?) hints take priority over path hints which can lead to fewer path hints than required dungeons (in extreme settings). Otherwise, it works great :p

PR #363, opened 2023-03-18
Starting interface choice

CovenEsme:
Adds a cosmetic option to set the in-game interface to Standard, Light, or Pro from the start of the game.

The asm is a bit strange since the rando doesn't currently have a tidy mechanism for setting startflags with a non-boolean value.


Muzugalium:
Tested standard, light, and pro interface options, all work

lepelog (approved):

PR #365, opened 2023-03-24
Path patches and Fire Node Block

elijah-thomas774:
Allows for patching paths
Should just need code review only. 
Fire node block was a suggestion to tie a permanent sceneflag to. A path needed to be created and added to accomplish this, thus this pr :) (finally lmao). 
Shouldnt need in game testing. I already verified the position.

Questions:
 I specified named parameters for the block patching, not sure if this is wanted
 idk where edge cases are when doing this patch, not familiar enough with python to know how it handles things not existing/wrong format (mainly talking about gamepatches.py)


patches.yaml
```diff
@@ -3972,6 +3972,24 @@ F300_3: # Fire node
...
     objtype: OBJ
+  - name: Add Path Idx and SceneFlag to Fire Node Box
+    type: objpatch
+    objtype: OBJ
+    id: 0xFC07
+    layer: 0
+    room: 0
+    object:
+      pathIdx: 1
+      goalscenefid: 48
```
lepelog, line 3983:
```suggestion
      goalscenefid: 127
```
That's the sceneflag of the fi trigger after tumbleweed chest, we delete the trigger so this sceneflag should be unused

lepelog (changes requested):
It kinda works, however the flag is only set in 2 positions, and moving the block anywhere else unsets the flag again, but that's probably not something we can easily fix?

elijah-thomas774:
Yeah sadly it can only be set in two positions. I dont think it is an issue for this block in particular, as it usually isnt touched after the first initial push down. I updated the sceneflag to the one you suggested

PR #368, opened 2023-03-26
use pdm

lepelog:
The formatting changes are just because of the black update

The rando version is now read from the pyproject.toml file, which means that `toml` is now a dependency


README.md
```diff
@@ -68,15 +68,17 @@ You will need Python (version 3.8 for Windows) and pip (which should come with P
...
+5. Install the dependencies:
+
+        pdm install
 
-5. [Run](#Running-the-randomizer) the randomizer:
+6. [Run](#Running-the-randomizer) the randomizer:
 
-        python randoscript.py
-    If your system uses old versions of Python, you may need to replace `python` with `python3`
+        pdm run randoscript.py randoscript.py
```
CovenEsme, line 81:
```suggestion
        pdm run randoscript.py
```

CovenEsme:
Does the `.pdm.toml` file that gets created after running `pdm` for the first time need adding to the gitignore?

lepelog:
Changes made

PR #369, opened 2023-03-27
Randomize Boss Key Puzzles

CovenEsme:
Aims:
* Randomize the starting angle of each boss key

Details:
* Adds an option to randomize the starting angle of each boss key
* Uses a new `bk_angle_seed`, similar to `trial_object_seed` and `music_rando_seed`
* Overwrites the vanilla asm that stores the bk angles with the random angles. Calls a custom function to read that data and write it to the `INITIAL_INSERT_ANGLE` array.

Tests needed:
* Check if the random bk angle works for ET, LMF and FS (I have tested the other three but more testing can't hurt)
* Verify having the option disabled keeps vanilla bk angles


CovenEsme:
Tested SV, AC, and FS boss keys last night in a co-op seed. Seems to be working so far :p

Muzugalium:
Tested in a few co-op seeds- works well

CovenEsme:
> Tested in a few co-op seeds- works well

Has now been tested with SV, LMF, AC, SSH, and FS. Just needs a test with ET and should be good :p

Muzugalium:
Tested boss keys in all dungeons, with and without random boss key puzzles turned on- works as expected

PR #379, opened 2023-04-04
prevent dying in thrill digger and bug heaven

lepelog:
To test:
- Have a damage multiplier > 2
- have more than 2 health, but as few as to die with one bomb
- go to bug heaven
- try to die to a bomb
- you should survive with 1 quarter heart

Repeat the process for thrill digger

Just to make sure I didn't mess up completely, try to die somewhere else to see that it still works


Muzugalium:
Tested both bug heaven and thrill digger and survived on a quarter heart in both

PR #380, opened 2023-04-05
Select Filters

cjs8487:
Adds the ability to add selection based filters to list pairs. This system requires metadata on the composition of the root list in order to function (specifically a mapping of displayed data to type list).

This change includes the addition of select filters for the excluded locations list pair, mimicking a toned down version of banned types from the previous versions of the randomizer. Most types were removed, leaving the following types which are useful either because they can't be text searched or they are an otherwise understandable grouping of checks.
- Goddess Chests
- Faron Woods Goddess Chests
- Lake Floria Goddess Chests
- Eldin Volcano Goddess Chests
- Volcano Summit Goddess Chests
- Lanayru Desert Goddess Chests
- Sand Sea Goddess Chests
- Combat
- Loose Crystals
- Minigames
- Beedle's Shop Ship Purchases
- Gratitude Crystal Sidequests
- Scrapper Deliveries
- Rupees

Marked as draft so initial review can begin, but note that filtering functionality is fragile and almost always immediately crashes

### DIFF
--- a/checks.yaml
+++ b/checks.yaml
@@ -103,7 +103,7 @@ Central Skyloft - Waterfall Cave Second Chest:
     - stage/D000/r0/l0/TBox/68
 Central Skyloft - Rupee Waterfall Cave Crawlspace:
   original item: "Red Rupee #25"
-  type:
+  type: Rupees
   Paths:
     - stage/D000/r0/l0/Item/88
 Central Skyloft - Parrow's Gift:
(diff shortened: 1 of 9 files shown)


CovenEsme:
Not sure what you have planned but it would be nice if there was a "Silent Realms" type.

Also, would it be possible to have the blank first entry in the types combo box have a default value to indicate that that option doesn't apply a filter?

lepelog:
Can we give the blank filter that shows all locations a better name? Maybe something like "All" or "Don't filter", I think that would make it easier

lepelog (approved):

PR #384, opened 2023-04-06
use poetry

lepelog:
let's hope there is dependency update support for this


CovenEsme (commented):
Works for me. Had an issue with getting `poetry run python randoscript.py` to work but it seemed to resolve itself.

Also, do we need setup.bat anymore? If yes, that should be updated as it still uses `requirements.txt`. Otherwise, it can be yeeted :p

PR #392, opened 2023-04-08
Allow Fire Dragon's Reward to be excluded

CovenEsme:
Fixes bug where you couldn't ban the Fire Dragon's Reward check. Similar to #329


Muzugalium:
Tested seeds generate with the fire dragon check both included and excluded- works

PR #393, opened 2023-04-08
slow bird down in levias fight

lepelog:
The range in which the cap is active around levias probably needs some fine tuning


asm/custom-functions/src/lib.rs
```diff
@@ -404,6 +433,33 @@ fn randomize_boss_key_start_pos(ptr: *mut u16, mut seed: u32) {
...
+#[no_mangle]
+fn enforce_loftwing_speed_cap(loftwing_ptr: *mut AcOBird) {
+    let loftwing = unsafe { &mut *loftwing_ptr };
+    let mut is_in_levias_fight = false;
+    if unsafe { &SPAWN_SLAVE.name[..4] } == b"F023" {
+        let levias_ptr = unsafe {
+            findActorByActorType(184 /* NusiB */, ptr::null())
+        };
+        if !levias_ptr.is_null() {
+            if unsafe { checkXZDistanceFromLink(levias_ptr, 5000f32) } {
```
CovenEsme, line 445:
```suggestion
            if unsafe { checkXZDistanceFromLink(levias_ptr, 20000f32) } {
```
The range is a bit small atm. Tried it with 20000 and it slowed down much further away - enough that I didn't accidentally crash into Levias.

CovenEsme (commented):
Works great. You slow down around the "saved" Levias as well as the "enemy" version. I know you mentioned making it only slow down during the fight so I figured I'd mention it. I don't think it's an issue tho. Means it's not a pain to land on Levias again if we ever add anything to the "saved" Levias in the future

lepelog:
changed so that it only slows down when the fight is active (based on storyflags)

CovenEsme:
Tested flying around the Thunderhead with scrapper holding the soup and after the Levias fight and there's no loftwing slowdown. During the fight, the loftwing speed noticeably reduces when within range of Levias.

Looks good to me :p

PR #401, opened 2023-04-20
Allow item get events to play underwater

CovenEsme:
Allows current event to be updated when getting an item. All items use the `DefaultGetItem` event when collected underwater.


lepelog:
included in #402

PR #402, opened 2023-04-25
Tadtonesanity

CovenEsme:
Aims:
* Add support for custom items
* Add 17 "Group of Tadtones" items to the pool
* Add 17 checks to Flooded Faron Woods (one for each vanilla tadtone group)
* Add option to toggle Tadtonesanity
* Allow entry into Flooded Great Tree without Water Scale
  * Add a bomb flower on the lilypad in-front of Bucha to prevent softlocks

Details:
* `keysanity.asm` has been renamed to `custom_items.asm`
* Custom functions have been added so that custom items can be more easily added in the future
  * This includes setting custom models for items and should be able to be used for adding the Sailcloth item (with its correct model)
* Logic for entering Flooded Great Tree has been updated to no longer require Water Scale
* As you need to watch the tadtones jingle in Flooded Faron Woods before getting the Water Dragon's Reward check, logic takes into account that you need access to Flooded Faron Woods in addition to Flooded Great Tree
* The Tadtone dowsing cutscene trigger has been removed since tadtone dowsing is a start flag. This can, ofc, be reversed if we ever want to make getting dowsing abilities checks
* The Cranioc cutscene trigger has been removed as it is buggy when getting it after already having gotten the Water Dragon's Reward check

Tests Needed:
* Verify getting a "Group of Tadtones" item from chests, npcs, freestanding, etc works as expected
* Verify the item get textbox shows the correct number of tadtones remaining
* Verify Fi shows you to have the correct number of tadtones
* Verify you can enter and leave Flooded Great Tree via the Water Dragon - even if you don't have Water Scale
* Verify you can get the Water Dragon's Reward check only after getting all 17 tadtones and that you can only get it once
  * You must watch the tadtones jingle in Flooded Faron Woods first (the one cutscene I won't remove xD)
* Verify you can get each tadtone check in Flooded Faron Woods both before getting the Water Dragon's Reward check and after
  * Getting all 17 should **not** start the tadtones jingle
  * The on-screen music scroll should update (even after getting the Water Dragon's Reward check)
  * The Cranoic and Tadtone dowsing cutscenes shouldn't play

Known Issues:
* The preview for the "Group of Tadtones" item in Beedle's Shop is raised higher than usual. This is true of other existing items and cannot be changed right now (not until someone looks into it at least)


checks.yaml
```diff
@@ -757,9 +757,155 @@ Lake Floria - Rupee on High Ledge outside Ancient Cistern Entrance:
...
+  type: Tadtones
+  Paths:
+    - stage/F103/r0/l2/Clef/0x019D
+    - stage/F103/r0/l2/Clef/0x019E
+    - stage/F103/r0/l2/Clef/0x019F
+    - stage/F103/r0/l2/Clef/0x01A0
+    - stage/F103/r0/l2/Clef/0x01A1
+    - stage/F103/r0/l2/Clef/0x01A2
+    - stage/F103/r0/l2/Clef/0x01A3
+    - stage/F103/r0/l2/Clef/0x01A4
```
lepelog, line 904: Is it possible to identify them by something like their sceneflag or group? Cause then we don't need to write out every tadtone

gamepatches.py
```diff
@@ -1221,6 +1226,18 @@ def rando_patch_goddess_crest(bzs: OrderedDict, itemid: int, index: str):
         obj["params2"] = mask_shift_set(obj["params2"], 0xFF, 0x18, itemid)
 
 
+def rando_patch_tadtone_group(bzs: OrderedDict, itemid: int, id: str):
+    id = int(id, 0)
+    clef = next(
+        filter(
+            lambda x: x["name"] == "Clef" and x["id"] == id,
```
lepelog, line 1233: I guess you *can* patch multiple Tadtones here, (if the group has something in common). If that works, I think it's better than having all the ids in checks.yaml

asm/patches/ss_necessary.asm
```diff
@@ -283,6 +283,56 @@ b 0x801b0764 ; return to original function
...
+rlwinm. r5, r5, 0x0, 0xd, 0xd ; is Link in water?
+cmpwi r5, 0
+bne 0x80251208 ; if in water, use DefaultGetItem event
+nop
+
+; Modify STORYFLAG_DEFINITIONS to allow flag 953 to be a counter
+; STATIC_STORYFLAGS[index] = 805a9b7e
+; shiftMask uses 7 lsb for the counter (max value of 128)
+; shiftMask >> 4 = 0 so:
+; Story Flag #953 (0x03B9) - US from 805A9B7E 0x01 to 805A9B7E 0x40
```
lepelog, line 312: what's at this address, just padding after the normal storyflags?

asm/custom-functions/src/lib.rs
```diff
@@ -367,6 +401,22 @@ fn rando_text_command_handler(_event_flow_mgr: *mut c_void, p_flow_element: *con
...
+            text_manager_set_num_args(&[storyflag_get_to_value(953) as u32]);
+        }
+        74 => { // Increment storyflag counter
+            let flag = flow_element.param1;
+            let increment = flow_element.param2;
+            
+            storyflag_set_to_value(flag, storyflag_get_to_value(flag) + increment);
+        }
+        75 => { // Have collected all tadtone groups?
+            let tadtone_groups_left = 17 - storyflag_get_to_value(953);
```
lepelog, line 415:
```suggestion
            let tadtone_groups_left: u32 = 17.saturating_sub(storyflag_get_to_value(953)).into();
```
Just to be safe, to not wrap around (side note, this only wraps around in release mode, in debug mode it would panic instead)
Also you can just use `into` here to get rid of the casts later, by just turning the u16 into u32 here

asm/custom-functions/src/lib.rs
```diff
@@ -90,6 +113,10 @@ fn dungeon_global_key_count(scene_index: u16) -> u16 {
     unsafe { (*dungeonflag_global(scene_index))[1] & 0xF }
 }
 
+fn storyflag_get_to_value(flag: u16) -> u16 {
```
lepelog, line 116:
```suggestion
fn storyflag_get_value(flag: u16) -> u16 {
```
or only `storyflag_get`, idk

lepelog (changes requested):
cool stuff :+1:

asm/patches/ss_necessary.asm
CovenEsme: Yeah. The normal storyflags start at index 0 and the counter storyflags start at the end and fill up space backwards. This uses the in-between space (after the last counter `Story Flag #1087 (0x043F) - US from 805A9B81 0x40 to 805A9B80 0x04`)

gamepatches.py
CovenEsme: This appears to work :p

CovenEsme:
Updated from feedback :p

gamepatches.py
```diff
@@ -1221,6 +1226,19 @@ def rando_patch_goddess_crest(bzs: OrderedDict, itemid: int, index: str):
...
+    groupId = int(groupId, 0)
+    clefs = filter(
+        lambda x: x["name"] == "Clef" and ((x["params1"] >> 3) & 0x1F) == groupId,
+        bzs["OBJ "],
+    )
+
+    next_clef = next(clefs, False)
+    while next_clef:
+        next_clef["anglez"] = mask_shift_set(next_clef["anglez"], 0xFFFF, 0, itemid)
+        next_clef = next(clefs, False)
```
lepelog, line 1239:
```suggestion
    for clef in clefs:
        clef["anglez"] = mask_shift_set(clef["anglez"], 0xFFFF, 0, itemid)
```

lepelog (approved):
looks good, only one nit and then this should be good to go

PR #405, opened 2023-04-29
Remove Ambiguous Options in Favor of Exclusions

cjs8487:
Removes the ambiguous options of Beedle's Shop Shuffle and Max Batreaux Reward in favor of using the exclusions system to prevent those locations from having progression..

## Changes
- Adds 4 new types to the excluded locations select filters
  - Batreaux's Rewards
  - Beedle's Shop Purchases (Cheap)
  - Beedle's Shop Purchases (Medium)
  - Beedle's Shop Purchases (Expensive)
- Removes the Max Batreaux Reward option completely
- Beedle's Shop option is now a boolean (checkbox) option for if the shop is shuffled or not

## Testing
- Mess with the UI, and ensure the filters behave as expected in various combinations
- Dry run a few seeds, no (new) errors should be produced
  - When the new Beedle's Shop option is disabled, all items Beedle sells should be vanilla

## Technical Note
- For now, I've elected to not make the Beedle option specific to Beedle internally, it is the generic "shopsanity" option. The display text is specific to Beedle just for simplicity and clarity. The help text is not generic, but does mention that Beedle is the only shop currently affected by the option.

### DIFF
--- a/SS Rando Logic - Glitched Requirements.yaml
+++ b/SS Rando Logic - Glitched Requirements.yaml
@@ -1097,17 +1097,17 @@ Can Access Beedle's Shop:
 Beedle's Shop - 300 Rupee Item:
   Can Access Beedle's Shop &
   Can Afford 300 Rupees &
-  (Progressive Pouch x1 | Option "shop-mode" Is Not "Vanilla")
+  (Progressive Pouch x1 | Option "shopsanity" Enabled)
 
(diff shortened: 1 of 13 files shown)


lepelog (changes requested):
Seems to work, you forgot to update the setting presets though

CovenEsme:
Given the Rupeesanity option a test spin. Seems good to me :p

PR #410, opened 2023-05-27
Various fixes (including an ugly patch)

yannl35133:


CovenEsme:
Starting with harp produces the attached error

image

yannl35133:
I fixed the issue for the second error, but couldn't reproduce the first one (I don't see how my patch would cause this in fact)

lepelog (approved):
can confirm that this does indeed fix the song nonprogress issue

PR #413, opened 2023-06-07
Dusk Relic Shuffling (For real this time)

YourAverageLink:
The old PR seems super borked so I'm opening this one from a new branch with significantly fewer commits.

Adds functionality to shuffle dusk relics in each silent realm (this is a continuation of an old Peppernicus PR)

Silent Realms can have up to 10 dusk relics turned into shuffled items. These items can swap locations with other trial objects with the trial object shuffle setting as well. Should you enable less than the maximum amount, the relics which contain items are chosen at random (the rest will have dusk relics as usual). Trial gates will remain open even after the trial has been completed (upon re-entry, you cannot obtain the trial reward again, nor will the treasure items respawn).

Song Hints still exist; they point specifically to the trial reward. Silent realms are now valid regions for other hint types. If song hints are disabled, trial rewards are now sometimes hints (rather than always hints).


SS Rando Logic - Glitched Requirements.yaml
```diff
@@ -2067,6 +2157,39 @@ Can Open Trial Gate in Lanayru Desert:
 Lanayru Silent Realm - Clawshots:
   Can Access Lanayru Silent Realm
 
+Lanayru Silent Realm - Clawshots:
+  Can Access Lanayru Silent Realm
```
CovenEsme, line 2161:
```suggestion
```
Duplicate entry

CovenEsme (commented):
Looks good to me. Testing seems to be going well since the Eldin trial bug was fixed. I have found it kinda difficult to distinguish the colours of the light beams on the items vs tears but I don't think there's anything we can do about that

gamepatches.py
```diff
@@ -1438,8 +1432,9 @@ def do_all_gamepatches(self):
         self.add_rando_hash()
         self.add_keysanity()
         self.add_demises()
-        if not self.placement_file.options["shuffle-trial-objects"] == "None":
-            self.shuffle_trial_objects()
+        self.shuffle_trial_objects()
+        # if self.placement_file.options["treasuresanity-in-silent-realms"]:
+        #     self.treasuresanity_in_silent_realms()
```
lepelog, line 1437: leftover

gamepatches.py
```diff
@@ -2504,19 +2500,27 @@ def shuffle_trial_objects(self):
...
+            types_to_shuffle = {"Tears", "Light Fruits", "Relics", "Stamina Fruits"}
+
         for trial in TRIAL_OBJECT_IDS:
             params = []
             locs = []
+            unshuffled_locs = []
+            relic_list = []
+            tear_list = []
+            lightfruit_list = []
+            stamina_list = []
```
lepelog, line 2521: are these actually used?

lepelog (changes requested):

goldendevil1711:
I may be dumb, but: is there a link to a build with just dusk relic shuffle (and not the other beta-features builds)?  I'm willing to test but given the but I had with beta-features yesterday I want to make sure I don't run into any weird conflicts.

CovenEsme:
Yep! Each PR has a build link near the top of the comment chain. You can find it here:

https://github.com/ssrando/ssrando/pull/413#issuecomment-1580024370

CovenEsme:
It'll automatically update to be the latest build too

YourAverageLink:
Please hold off on merging this until the weekly is over (so the version and settings string are still up to date and the tracker works correctly).

PR #415, opened 2023-06-10
actually fix firedragon when both other dragons are met

lepelog:
my previous fix was incomplete, but now it actually directly patches the fire dragon even, so there is no chance of it going to the unpatched flow which doesn't give the item and spits you out near earth temple


CovenEsme:
Verified the incorrect behaviour and tested this fix. Works and looks good to me :p

PR #418, opened 2023-06-14
Fix patching bug when starting with all tadtones

CovenEsme:
Does what it says on the tin. My smallest ever change xDDD


CovenEsme:
Spoke too soon, the asm needed changing too xD

PR #428, opened 2023-06-22
Reduced Peatrice Convos

CovenEsme:
Adds an option to reduce the number of Peatrice conversations needed before she calls you "Darling".

Text for each value:
* 6 = default (no change)
* 5 = "...Oh, it's you again..."
* 4 = "But I bet you're here just to check me out, right?"
* 3 = "Oh. I, uhh... W-welcome..." ("What's wrong?" vs "You seem different." dialog choice)
* 2 = "Oh. You're back."
* 1 = "...Oh." (calls you "Darling" at the end)
* 0 = "Hello, darling!" (recurring dialog until you talk to Peater)


asm/custom-functions/src/lib.rs
```diff
@@ -305,6 +305,31 @@ pub fn process_startflags() {
 
     let mut pouch_slot_iter = unsafe { (*FILE_MANAGER).FA.pouch_items.iter_mut() };
 
+    // Peatrice convo count.
```
lepelog, line 308: If you ask me, these startstoryflags should be set like all others in `gamepatches.py`, with 7 flags max that's only 14 bytes, or are we that low on space for startflags already?

CovenEsme: Yeah, that's fair. I just know we're pushing the limit already (starting with all items puts us 6 bytes over). I've been trying to keep the number of startflags down to a minimum until I get round to changing how they're processed.

I initially tried putting them in `patches.yaml` but the logic was a mess to read. Didn't consider putting them in `gamepatches` tho

CovenEsme: Done :p

CovenEsme:
Tested in a seed where Peatrice was required (SotH piece) with 0 convos required. Worked as expected

PR #431, opened 2023-06-24
Counters-re

yannl35133:


YourAverageLink:
Lgtm. Generated some seeds, they took fewer years to generate and I haven't noticed any logical oddities yet. Big fan of the move away from EINs for logic options.

YourAverageLink:
Actually upon further inspection it looks like Beedle's shop now has some inlined requirements. This isn't as much of a big deal as, for example, Dodoh's Crystals or Fire Dragon's Reward from before, but it's still something to note.

yannl35133:
> Big fan of the move away from EINs for logic options.

This might be a move in the direction opposite to what you think. Sure they get inlined at startup instead of existing during the whole randomisation process, but they still exist as an index to a requirement at some point (although not in the usual `requirements` array).

> Actually upon further inspection it looks like Beedle's shop now has some inlined requirements.

Not inlined requirements, but other checks as requirements (not the same thing)

logic/logic_expression.py
```diff
@@ -261,25 +628,139 @@ def mk_and(self, left, right):
...
+    ?int_expr:
+        | INT -> mk_int
+        | "Option" text -> mk_int_option
+        | "Option" text "{" (assoc_pair ("," assoc_pair)*) "}" -> mk_any_option
+
+    ?assoc_pair: text "->" INT -> mk_pair
+
+    ?counter: counter "+" counter -> mk_counter_add
+        | INT ("x" | "*") TEXT -> mk_counter_atom1
+        | TEXT "{" (INT ("," INT)*) "}" -> mk_counter_atom2
```
lepelog, line 707: I think this is overengineered. If I see it correctly, this is only used in 3 instances (2 is not counting `Unique Rupees`, which shouldn't be logically relevant).

Doing this in plain python code would make it much easier to read and maintain.

Please don't do this, but if you want this special syntax to be used you would need to properly document it. `mk_counter_atom2` doesn't speak for itself.

PR #433, opened 2023-06-26
pin rust toolchain and reassemble

lepelog:


CovenEsme:
Tested this when merging it into `beta-features` and it updated my version and assembled the stuff. Not sure if anything else needs testing

PR #434, opened 2023-06-26
check that yaml keys are unique

lepelog:
also removed the obsolete `file_accessor`, `yaml_files` does the same thing basically


CovenEsme:
Looks good to me. Caught the duplicate entry in #413

PR #435, opened 2023-06-27
Scrapper & Gondo

CovenEsme:
Adds Scrapper as an item needed to do all Scrapper Quests. Scrapper had also been added as a possible start item.

Adds a check to Gondo: `Central Skyloft - Repair Gondo's Junk`
The check requires 1 Ancient Flower.

Testing:
* Ensure the Crystal Ball, Oolo, Party Wheel, and Soup cannot be collected without the Scrapper item (I've tested Party Wheel and Soup)
* Test if you can collect a scrapper item after just obtaining Scrapper without a reload
* Test different combos of upgrading items, repairing shields, and getting the Gondo check. It should all be good but wanna be sure, ya know :p
* Look over logic and see if I've screwed anything up :p


CovenEsme:
I've tested this in the solo async (special) and in another seed I've done :p

yannl35133:
Always spawn ancient flowers ?

CovenEsme:
Yeah, there's already been feedback from the asyncs that ppl put off getting the "Repair Gondo's Junk" check because a flower didn't spawn for them. It's only a matter of time before someone loses a race as a result

cjs8487:
I’d much rather seeding in game rng as a long term solution. In the interim, maybe guaranteeing the first spawn in mines specifically could work as a good solution? Not a huge change, still requires Lanayru access, and maintains normal treasure spawning rules elsewhere (which matters for things like bingo)

CovenEsme:
Reverting the change as it can be dealt with later

CovenEsme:
Tested with a full seed playthrough, no issues found

PR #438, opened 2023-06-29
catch and print errors during bulk generation

lepelog:


CovenEsme:
Tested and seems to work :p

PR #440, opened 2023-07-03
yeet wit and use a terribly written library instead

lepelog:


CovenEsme:
Tested with a full seed playthrough and worked well. Had to rename the iso to `SOUE01.iso` for usbloader gx / wiiflow to actually start the game tho

extractmanager.py
```diff
@@ -0,0 +1,105 @@
...
+        if not self.actual_extract_already_exists():
+            dest_path = self.rootpath / "actual-extract"
+            extractor = disc_riider_py.WiiIsoExtractor(iso_path)
+            extractor.prepare_extract_section("DATA")
+            # remove hint videos, but keep credits
+            extractor.remove_files_by_callback(
+                "DATA", lambda x: x.startswith("THP") and not "Demo" in x
+            )
+            checksum = bytes(extractor.get_dol_hash("DATA")).hex()
+            if "450a6806f46d59dcf8278db08e06f94865a4b18a" != checksum:
```
CovenEsme, line 50:
```suggestion
            if CLEAN_NTSC_U_1_00_DOL_HASH != checksum:
```
You declare the constant but don't actually use it :p

CovenEsme:
Not sure how much testing we want before this gets merged but this works for several ppl and looks good to me. The only thing I'd say is that the README should probably be updated to refer to a .iso file rather than a .wbfs

PR #442, opened 2023-07-05
Bizzare Bazaar

CovenEsme:
Adds 2 new forms of shopsanity: Gear Shopsanity and Potion Shopsanity. When enabled, you can buy a randomized item (on time only) and then you can buy the vanilla item repeatedly.

All of Rupin's items are available to purchase immediately. Luv's items require a bottle and a pouch (this check had to be added in). For the stamina potion and air potion items, you also need to have raised LMF or obtained the water scale.

The shopsanity patches are now always enabled (except for the potion shop specific ones).

Tests Needed:
* Test I didn't break anything with Beedle. I haven't come across anything but 🤷‍♀️ 
* Make sure you can only get each item once and that you can repeatedly get the vanilla item afterwards
* Test different combos of shops on and off to make sure they don't interact weirdly


CovenEsme:
Tested with a full seed playthrough, no issues found

goldendevil1711:
Tested this today on beta-features with full Bazaar as well as full Beedle.  Didn't have any issues with it, no errors that I could see.

goldendevil1711:
One suggestion: it would be very helpful to put the prices into the check names for both the randomizer and the tracker.  Right now, it's impossible to know what the prices are for every check.

CovenEsme:
I am not currently interested in updating this. If anyone is interested in picking this up is welcome to do so.

Please note that #463 has not been resolved

PR #445, opened 2023-07-08
fix placement file

### LINKED ISSUE #444: Python error on Linux when trying to use a plando file
SSRando `v1.4.1` git tag; Kubuntu 23.04; Python 3.11.2
`./ssrando.sh --placement-file /path/to/my.ssplando.json`
```
Traceback (most recent call last):
  File "/[...]/ssrando/randoscript.py", line 189, in <module>
    main()
  File "/[...]/ssrando/randoscript.py", line 120, in main
    plcmt_file.check_valid()
TypeError: PlacementFile.check_valid() missing 1 required positional argument: 'areas'
```
It's saying [check_valid](https://github.com/ssrando/ssrando/blob/v1.4.1/logic/placement_file.py#L69) method is not being passed a required `areas` argument. So it doesn't seem to be a problem with my plando file I think, since the method really is [being called](https://github.com/ssrando/ssrando/blob/v1.4.1/randoscript.py#L120) with no arguments other than self: `plcmt_file.check_valid()`.

Or am I using this placement file or command line wrong? How should I use it?
I copied it from a previous generation from the GUI, but removed most key/value pairs except top-level keys, and adjusted the things I wanted.

lepelog:
fixes #444

### DIFF
--- a/randoscript.py
+++ b/randoscript.py
@@ -111,14 +111,16 @@ def main():
             print(err)
         exit(1)
 
+    areas = Areas(requirements, checks, hints, map_exits)
+
     plcmt_file_name = parsed_args.placement_file
     if plcmt_file_name is not None:
         plcmt_file = PlacementFile()
         with open(plcmt_file_name) as f:
             plcmt_file.read_from_file(f)
-        plcmt_file.check_valid()
+        plcmt_file.check_valid(areas)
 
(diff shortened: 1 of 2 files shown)


yannl35133 (approved):
Assuming you tested, lgtm

PR #447, opened 2023-07-10
Prevent flows from setting storyflag -1

YourAverageLink:
Fixes a bug found in #413 and probably any other use of storyflag -1 (2047) as 'no storyflag'. Tested the unrequired FS final item flow which sets both a real storyflag and tries to set storyflag 2047. Worked as intended, setting only the real storyflag.


asm/custom-functions/src/lib.rs
```diff
@@ -592,6 +588,11 @@ fn give_item_with_sceneflag(item_id: u16, bottle_pouch_slot: u32, number: u32, s
     }
 }
 
+#[no_mangle]
```
cjs8487, line 591: Not an actual issue, but is there a practical reason that this needed to be moved, rather than just adding the `#[no_mangle]`?

YourAverageLink: Not really functional, just so it's with all the rest of the no-mangle functions (there's a comment at the top of the no mangle list explaining why it's used so I felt it was appropriate)

asm/patches/ss_necessary.asm
```diff
@@ -333,6 +333,13 @@ nop
 nop
 nop
 
+; don't allow setting storyflag -1 (0x7FFF) in flows (breaks trials & probably other things)
+.org 0x801aff50
+cmplwi r4, -1
+beq 0x801aff60
```
lepelog, line 339:
```suggestion
cmplwi r4, 2047
bge 0x801aff60
```
otherwise I don't see how 2047 is prevented from being set, this would also deny all other "negative" numbers

YourAverageLink: I should specify, 0xFFFF is in register 4 when this comes up in the assembly if the storyflag is -1

lepelog: 2047 is 0x7FFF though

YourAverageLink: I dunno, the patch seems to correctly block the -1 in the unrequired dungeon flow. Maybe the storyflag manager &s the value with 7FF.

CovenEsme: when logically compared with `cmplwi`, -1 -> 0xffff not 0x7fff

lepelog: -1 != 2047, I know this patch works for -1, but I'm saying it doesn't work for 2047. But I checked and no flows reference 2047, only -1, so we can keep it as it is right now

lepelog (approved):

PR #449, opened 2023-07-12
Fix crystal quests not appearing under the correct select filter

cjs8487:
Does what it says on the tin


CovenEsme:
Should Dodoh's Crystals and FFI also have the "Gratitude Crystal Sidequests" check type?

cjs8487:
I didn't change which things had the types, just the typo. Not sure why Dodoh doesn't, although I think FFI doesn't because we wanted to avoid the transitive typing that we had in the old system. Up for changing it though

PR #458, opened 2023-07-22
Place restricted items first

CovenEsme:
Prevents those `No more locations accessible for item` errors from occurring by placing items with placement restrictions before all other items.


CovenEsme:
Closing as I don't remember what I was doing with this ^^'

PR #460, opened 2023-07-31
Fix heart spawning issue

CovenEsme:
Fixes an issue where hearts wouldn't spawn from digspots, pots, most enemies, and TgReacts even when the option was enabled


CovenEsme:
If no one's got any issues with this, I'll merge it in a couple days

goldendevil1711:
Not sure if anyone confirmed but I tested this today on beta-features and there were many more spawns of hearts from enemies than in the prior version.

PR #461, opened 2023-08-01
Prevent crash from dying without reload

CovenEsme:
Should truly fix the crash when dying on a new game file without a reload. Shouldn't have any weird BiT or EAT side effects and all for the cost of one storyflag


CovenEsme:
With the global var, I've tested:
* dying without reload (no crash)
* using a corrupt file for BiT (works)
* the EAT stuff (doesn't happen)

So all the previously known issues with this fix seem to be non-issues now

PR #464, opened 2023-08-06
Make locations in an already barren-hinted region unhintable

YourAverageLink:
This functionality was brought up a while ago but shot down for fear of removing useful dungeon sometimes hints for dungeon boss keys. In my opinion, that case (a dungeon being barren at all + an otherwise useless location hint that would have been placed) is so rare now that it's a non-issue (plus, if this behavior isn't desired, it can be prevented if all sometimes hints come before barren in the hint distribution. I believe it is preferable to prevent otherwise useless sometimes hints from being placed.

The 2D Dowsing & Fi Hints distribution has also been updated to place barren hints before sometimes -- this distro was especially vulnerable to useless sometimes hints given how relatively weak the hint pool is.

### DIFF
--- a/hints/distributions/2D Dowsing & Fi Hints.json
+++ b/hints/distributions/2D Dowsing & Fi Hints.json
@@ -34,10 +34,10 @@
   "dungeon_barren_limit": 1,
   "distribution": {
     "always": {"order":  0, "weight":  0.00, "fixed":  0, "copies":  1},
-    "sometimes": {"order":  1, "weight":  1.20, "fixed":  4, "copies":  1},
-    "goal": {"order":  2, "weight":  0.00, "fixed":  1, "copies":  1},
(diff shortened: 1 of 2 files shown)


yannl35133:
There is already an opt-in mechanism for hints to see if the location they hint is in a barren area, why did you not add it for sometimes hints ?

YourAverageLink:
> There is already an opt-in mechanism for hints to see if the location they hint is in a barren area, why did you not add it for sometimes hints ?

I wanted it to only prevent the hint being placed if the corresponding barren hint has already been placed. For example, if there were already 3 barren hints placed but summit is also barren (and wasn't hinted as such), I want it to still be possible to place a sometimes hint for the Item behind Digging

yannl35133:
I was unclear, I meant that the mechanism allows to check if the location is in a *hinted-barren area*

PR #469, opened 2023-08-31
refactor startflags to be less constrained for space

lepelog:
Adding more story- and itemflags in the future won't make a difference size wise. For sceneflags there is still a limit, but that's like 60 flags, so we see if we ever get there.

Also includes some slight changes to the assemble script, so that unused stuff the compiler sometimes generates is removed.

This also fixes a bug where some bugs and treasures wouldn't show up when starting with them.


util/flag_mapping_tables.py
```diff
@@ -0,0 +1,1791 @@
...
+#     if value > mask:
+#         raise Exception(f"flag {flag}, value {value} is greater than allowed {mask}")
+
+#     buffer[index] = (buffer[index] & ~(mask << shift)) | (value << shift)
+
+# def set_storyflag(buffer: List[int], flag: int, value: int):
+#     _set_flag(storyflag_mapping_table, buffer, flag, value)
+
+# def set_itemflag(buffer: List[int], flag: int, value: int):
+#     _set_flag(itemflag_mapping_table, buffer, flag, value)
```
CovenEsme, line 1762:
```suggestion
```

Leftover?

util/flag_mapping_tables.py
```diff
@@ -0,0 +1,1791 @@
...
+
+
+class MappedFlagWriter:
+    def __init__(self, length: int, mapping_table: List[int]):
+        self.buffer = [0] * length
+        self.mapping_table = mapping_table
+
+    def set_flag(self, flag: int, value: int):
+        definition = self.mapping_table[flag]
+        if definition == 0xFF00:
```
CovenEsme, line 1772: Might be worth defining an `UNDEFINED_FLAG` constant and using that. Could make it easier to verify the mapping tables are correct

CovenEsme (commented):
Looks good to me. Just a few commented out bits of code that could be yeeted and a couple of questions/observations

asm/custom-functions/src/lib.rs
```diff
@@ -237,62 +237,38 @@ fn itemflag_set_to_value(flag: u16, value: u16) {
...
+        STATIC_ITEMFLAGS = startflag_info.itemflags;
     }
     // sceneflags
-    while let Some(flag) = flag_mem.next_u16() {
-        if flag == 0xFFFF {
+    let mut scene_idx = 0;
+    for flag in startflag_info.sceneflags.iter() {
+        if *flag == 0xFF {
             break;
+        } else if *flag >= 0x80 {
```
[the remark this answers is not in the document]
yannl35133: (I haven't read the rest of the code, I may be missing some context)
`flag` is defined by the `for` loop, and since this is an iterator it gets assigned to the different references the iterator holds.
To use the reference (for comparisons), you have to dereference it

util/flag_mapping_tables.py
lepelog: I've now reformatted the tables so they take up less space, so it doesn't make sense to use such a constant in them anymore. Do you think it makes sense to still make it a constant for this use here? I usually don't create constants for single use values, but I don't really mind

CovenEsme: Na, this is fine :p

CovenEsme (approved):

PR #474, opened 2023-09-10
Add support for conditional hints

CovenEsme:
**TODO**: If someone could go through the distros and update/remove them to bring them up-to-date with, that would be hugely appreciated (don't feel shy to just edit this PR or PR onto it :p)

Does what it says on the tin. I did this a couple months back so I don't remember much of what I did.

I've adjusted the "balanced" distro to take full advantage of the conditional hints and should be a good baseline.

I removed the S2 hint distros based on discussion we had in discord a while back?

The remaining distros have been updated so that they don't error but otherwise haven't been changed


CovenEsme:
Oh, and apparently the triforces, sea chart, and the key pieces have been added as possible items to hint if they are relevant to beating the seed :p

YourAverageLink:
I'm pretty sure sea chart was already hintable, but that's nice for the other two

CovenEsme:
I'm not really sure. I think the ppl that are currently more actively interested in reworking how the hints work could give a better answer.

The goal with this PR was to basically reduce the need for multiple hint distros to only be for tweaking weightings. The conditional side was meant to allow us to determine under what conditions any given check should be hintable based on the enabled settings. I think this can be achieved with the merged PR? Again, Robo or Golden would be able to give a more definitive answer

CovenEsme:
Closing as it's just a different design for the same thing that #568 added. No reason to keep this PR open afaik

PR #475, opened 2023-09-10
Rename DER option

CovenEsme:
Renames the "All Dungeons" DER setting to "All Surface Dungeons" to better describe its behaviour

### DIFF
--- a/gui/presets/default_presets.json
+++ b/gui/presets/default_presets.json
@@ -124,7 +124,7 @@
         "triforce-required": true,
         "triforce-shuffle": "Anywhere",
         "seed": -1,
-        "randomize-entrances": "All Dungeons + Sky Keep",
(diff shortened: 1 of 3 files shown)


cjs8487 (commented):
Should we also rename "All Dungeons + Sky Keep" -> "All Dungeons"

CovenEsme:
I thought about that but since we don't treat Sky Keep as a proper dungeon (like, it can't be one of the required dungeons), I think it's clearer to keep that as is?

cjs8487:
Yeah that's a fair point. If we want to keep " + Sky Keep" maybe we should change it to "All Surface Dungeons + Sky Keep"? Makes it a bit clearer that the only change is adding Sky Keep into the mix, but I don't really mind either way

CovenEsme:
Cool, works for me

CovenEsme:
Will require a tracker update when this gets merged :p

PR #479, opened 2023-09-27
Fix SSH + DER wrong dungeon exit bug

CovenEsme:
When beating SSH with Practice Sword, Fi warps you out of the dungeon. When returning with a better sword to strike the crest, you would be taken to the docks in the Ancient Harbour. This fixes that bug


CovenEsme:
I've already tested this, just gonna wait for the gh action to successfully complete and I'll merge it

PR #481, opened 2023-09-27
Change to trick for Lightning Node raised chest

CovenEsme:


yannl35133:
You still need the bomb bag to do this

CovenEsme:
Oh, so it should be `Beetle | Bow | (Lightning Node End with Bombs Trick & Bomb Bag)`?

yannl35133:
Yes (parentheses only needed because you want them)

PR #483, opened 2023-09-27
Add AC rupee under lilypad check

CovenEsme:
Adds the `Ancient Cistern - Rupee under Lilypad` check. It was previously thought that you could lock yourself out of this check but this is not true. However, I have moved the rupee further into the underwater tunnel to prevent it from getting caught up in the water stream (which made it very hard to see).

I've posted a video of how to get the check with the water stream already active in disord


CovenEsme:
Tested with a glitchless AC full-clear and all was fine. I'll merge this if nobody complains :p

PR #484, opened 2023-09-27
Add call to the rust formatter into assemble.py

CovenEsme:
I've added `rustfmt.toml` too. Feel free to change the config options to suit your style, Lepe - I just had a quick look through to see what was doable


asm/assemble.py
```diff
@@ -430,13 +430,17 @@ def free_space_org_list_sorter(code_chunk_tuple):
                     "-o",
                     bin_name,
                 ]
+
+                if result := call(["cargo", "fmt"], cwd="./custom-functions"):
+                    raise Exception("Formatting rust functions failed.")
+
                 # add custom functions from rust
                 if is_custom_function and file_path == "main.dol":
```
lepelog, line 438:
```suggestion
                # add custom functions from rust
                if is_custom_function and file_path == "main.dol":
                    if result := call(["cargo", "fmt"], cwd="./custom-functions"):
                        raise Exception("Formatting rust functions failed.")
```
Otherwise it formats for *every* asm patch, which is unnecessary

lepelog (changes requested):
format config looks fine to me :+1:

asm/assemble.py
CovenEsme: Ah yeah, good catch

lepelog (approved):

PR #490, opened 2023-10-02
Fix hint distros

CovenEsme:
* Fixes the location names in the `Boss Keysanity Fi Hints` hint distro
* Adds code to make hint distros resilient to incorrect location names
  * If a distro has an incorrect name, it will print a warning to the console but still generate the seed


hints/hint_distribution.py
```diff
@@ -177,21 +177,26 @@ def start(
...
+                self._get_full_location_name(location["location"]): location["type"]
+                for location in self.added_locations
+            }
+            | {
+                self._get_full_location_name(location): None
+                for location in self.removed_locations
             }
-            | {areas.short_to_full(loc): None for loc in self.removed_locations}
         )
         # Combines those 3 dictionaries, the right-most dict has priority when keys are shared
```
yannl35133, line 191: Don't try to return the short form location, there is zero chance it will work. Either raise an error or explicitly ignore it.
Also, I don't see the point in making this helper function global
```python
def wrap_stf(location: str) -> EXTENDED_ITEM_NAME:
    try:
        return self.areas.short_to_full(location)
    except ValueError:
        # Still continue generating the seed if the distro isn't 100% correct
        print(f"Could not find location with name: {location}, entry will be ignored")
    return None

hint_status_from_distro = (
    check_hint_status
    | {
        wrap_stf(loc["location"]): loc["type"]
        for loc in self.added_locations
    }
    | {wrap_stf(loc): None for loc in self.removed_locations}
)
# Combines those 3 dictionaries, the right-most dict has priority when keys are shared

hint_status_from_distro.pop(None, None)
# Remove the ignored entries, if any
```

CovenEsme: Would have implemented this but I'm removing this until further discussion since the hint distros are getting reworked soon:tm: anyways. Thanks for the review tho

PR #494, opened 2023-10-11
show small key count when getting key

lepelog:


CovenEsme:
Tested with Skyview Small Keys on Fledge's Gift and Owlan's Gift. Tested getting both keys and the counts were correct. I also tested getting Fledge's key, going to Skyview and using it, coming back and getting Owlan's key. That worked too :p

PR #504, opened 2023-11-24
Logic dump

cjs8487:
Adds dump mode and an automated action to check if the dump file contents in the repository match the contents generated by the randomizer. Failing the dump check does not prevent builds from completing. Dump mode dumps to yaml format.


randoscript.py
```diff
@@ -111,6 +118,18 @@ def main():
...
+        logic.logic_expression.GLOBAL_DUMP_MODE = True
+        areas = Areas(requirements, checks, hints, map_exits)
+        if dest is True:
+            print(areas)
+            exit(0)
+        with open(dest, mode="w") as f:
+            print(yaml.dump(areas.to_dict()), file=f)
+            exit(0)
+
     areas = Areas(requirements, checks, hints, map_exits)
```
YourAverageLink, line 133: I guess this was here from NULL's original branch, but can't this be moved before the if block (and then line 125 can be deleted)?

cjs8487:
Should be ready for review now - actively testing how valuable the dump contents actually are in a downstream consumer

lepelog:
obsoleted by #513

PR #510, opened 2023-12-13
Allow specific hint types for removed items

YourAverageLink:
As the title says.

Allows control over whether items are removed for sots/path hints or item hints.


CovenEsme:
Looks good to me

cjs8487 (approved):
Code looks good, will merge when we get confirmation of retesting

PR #512, opened 2023-12-24
Fix some duplicated locations

robojumper:
These were found by a simple "unused location" check in the new-logic tracker (area locations that don't correspond to a check and also aren't mentioned in any other expressions). This fixes the ones that very obviously look like bugs -- there are more unused locations but some of them are simply unused in a benign way.

### DIFF
--- a/logic/requirements/Eldin.yaml
+++ b/logic/requirements/Eldin.yaml
@@ -59,9 +59,7 @@ Volcano:
     locations:
       Unlock Shortcut to Pre Turf: Nothing
       Watch Bridge Cutscene: Nothing
-      East - Southeast Rupee above Mogma Turf Entrance: Beetle
-      East - North Rupee above Mogma Turf Entrance: Beetle
-      East - Goddess Cube near Mogma Turf Entrance: Goddess Sword
+      # can get cube and mogma turf entrance rupees from here, but getting to East from here requires nothing
 
   Ascent:
     exits:
@@ -326,7 +324,7 @@ Bokoblin Base:
       First Chest in Volcano Summit: Fireshield Earrings
       Raised Chest in Volcano Summit: Fireshield Earrings
       Chest in Volcano Summit Alcove: Fireshield Earrings & Practice Sword
-      Volcano Summit - Goddess Cube inside Volcano Summit: Fireshield Earrings
+      Volcano Summit - Goddess Cube inside Volcano Summit: Fireshield Earrings & Goddess Sword
   
   Fire Dragon's Lair:
     exits:
--- a/logic/requirements/Eldin.yaml
+++ b/logic/requirements/Eldin.yaml
@@ -59,9 +59,9 @@ Volcano:
     locations:
       Unlock Shortcut to Pre Turf: Nothing
       Watch Bridge Cutscene: Nothing
-      Southeast Rupee above Mogma Turf Entrance: Beetle
-      North Rupee above Mogma Turf Entrance: Beetle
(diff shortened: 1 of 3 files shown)


logic/requirements/Eldin.yaml
```diff
@@ -59,9 +59,9 @@ Volcano:
     locations:
       Unlock Shortcut to Pre Turf: Nothing
       Watch Bridge Cutscene: Nothing
-      Southeast Rupee above Mogma Turf Entrance: Beetle
-      North Rupee above Mogma Turf Entrance: Beetle
-      Goddess Cube near Mogma Turf Entrance: Goddess Sword
+      East - Southeast Rupee above Mogma Turf Entrance: Beetle
+      East - North Rupee above Mogma Turf Entrance: Beetle
+      East - Goddess Cube near Mogma Turf Entrance: Goddess Sword
```
robojumper, line 64: These are redundant since just blowing up the rock with the bomb flower ("Nothing") is enough to get to East, so they could be dropped too.

logic/requirements/Eldin.yaml
```diff
@@ -273,7 +273,7 @@ Bokoblin Base:
     locations:
       Chest near Bone Bridge: Can bypass Boko Base Watch Tower | Mogma Mitts | Gust Bellows
       Chest on Cliff: (Can bypass Boko Base Watch Tower | Mogma Mitts) & Gust Bellows
-      Goddess Cube near Mogma Turf Entrance: Goddess Sword
+      Volcano - Goddess Cube near Mogma Turf Entrance: Goddess Sword
```
robojumper, line 276: I assume you can actually get these in Boko Base, I haven't checked.

yannl35133:
I suspect the "benign" unused locations are hooks for BiT logic in the future, but these definitely look like errors

logic/requirements/Lanayru.yaml
```diff
@@ -56,7 +56,7 @@ Mine:
       exits:
         Exit to Desert: Nothing
       locations:
-        Blow up Rock on Track: Bomb Bag & Lanayru Mine - Quick Bomb Trick
+        End - Blow up Rock on Track: Bomb Bag & Lanayru Mine - Quick Bomb Trick
```
robojumper, line 59: If I'm understanding this right I think "In Minecart" is also missing an exit to "End" predicated on being able to blow up the rock so that logic actually allows you to leave the minecart into End once the rock is blown up using this trick?

lepelog: Yea, but as long as blowing the rock up from the minecart is not even a trick, it wouldn't change logic at all so it's omitted

logic/requirements/Eldin.yaml
```diff
@@ -326,7 +326,7 @@ Bokoblin Base:
       First Chest in Volcano Summit: Fireshield Earrings
       Raised Chest in Volcano Summit: Fireshield Earrings
       Chest in Volcano Summit Alcove: Fireshield Earrings & Practice Sword
-      Goddess Cube inside Volcano Summit: Fireshield Earrings
+      Volcano Summit - Goddess Cube inside Volcano Summit: Fireshield Earrings
```
lepelog, line 329:
```suggestion
      Volcano Summit - Goddess Cube inside Volcano Summit: Fireshield Earrings & Goddess Sword
```

logic/requirements/Eldin.yaml
lepelog: I think they should be removed then

lepelog (changes requested):

PR #513, opened 2023-12-25
Logic dump v2

robojumper:
A rebase and continuation of #504.

> Adds dump mode and an automated action to check if the dump file contents in the repository match the contents generated by the randomizer. Failing the dump check does not prevent builds from completing. Dump mode dumps to yaml format.

This dump is currently being used by the [WIP new-logic tracker](https://robojumper.github.io/SS-Randomizer-Tracker/) and it seems to be reasonably complete (with some hardcoding needed for special requirements and logical rules).

The main concern for the future is compatibility - mainly, should this dump format be able to change in an incompatible way due to rando changes or evolutions of the dump format, and if so, when? Should we emit a "spec version" in the dump so that consumers can handle breaking changes? Should the eventual release of the new-logic tracker only support the latest version of the dump, with legacy versions of the tracker for older rando versions with incompatible dumps hosted separately? Until these questions are resolved, I think it's fine to advertise this dump as unstable and likely to change in the near future.

### DIFF
--- a/.github/workflows/dump-check.yaml
+++ b/.github/workflows/dump-check.yaml
@@ -20,6 +20,6 @@ jobs:
       - name: Install system dependencies
         run: sudo apt-get update && sudo apt-get install ffmpeg libsm6 libxext6  -y
       - name: Create dump file
-        run: python3 randoscript.py --noui --dump dump-check.yaml
+        run: PYTHONHASHSEED=0 python3 randoscript.py --noui --dump dump-check.yaml
       - name: Check contents
         run: diff dump.yaml dump-check.yaml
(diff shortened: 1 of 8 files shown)


lepelog:
I don't think we need to version the dump format, the only change I can see in the future would be adding more fields, which should be backwards compatible.

We should document how to create the dump format though, so people who edit the logic files know how to do it

CovenEsme:
Pretty sure this is the same issue that we have with asm changes. If regenerating the logic dump is trivial (I haven't been following this PR closely), I wouldn't block a PR on this issue. Would be nice if merge queues helps smooth the process tho :p

lepelog:
I don't think we have to worry about semantic merge conflicts here, even if we somehow break main we can quickly fix it. I would expect 2 PRs that modify logic to produce a merge conflict in the dump file, so I would expect it's not an issue anyways

PR #517, opened 2023-12-29
Print to screen fmt

lepelog:


lepelog:
already implemented in #526

PR #523, opened 2024-01-12
Prevent picking up Item behind Bars with Sword

CovenEsme:
Fixes the newly discovered bug that allowed you to pick up the `Skyview Temple - Item behind Bars` item with the Sword.

I've tested:
* you could pick up the item with a Skyward Spin Attack before the patch
* you can't pick up the item in the same way after the patch
* you can still whip the item through the bars
* you can still get the item by just running into it


lepelog:
The code is a bit weirdly structured there, but if I see it correctly this patch only affects the freestanding item in skyview, so should be good

PR #525, opened 2024-01-14
Random Starting Statues

YourAverageLink:
Adds an option to have a random bird statue to start with in each region, rather than the vanilla entrance statues. This excludes the Inside the Volcano and Inside Fire Sanctuary statues, as they would leave the player softlocked.


gamepatches.py
```diff
@@ -3286,3 +3319,16 @@ def do_patch_custom_dowsing_images(self):
         ).read_bytes()
         arc.set_file_data("timg/tr_dauzTarget_18.tpl", sandshipdata)
         do_button_path.write_bytes(arc.to_buffer())
+
+    def patch_random_starting_statue_flags(self):
+        for statue in self.placement_file.start_statues.values():
+            flag = BIRD_STATUE_FLAGS.get(statue[1], None)
```
lepelog, line 3325:
```suggestion
            flag = BIRD_STATUE_FLAGS.get(statue[1])
```
The None is redundant

changedflags.md
```diff
@@ -84,6 +84,9 @@ New Trial Completed Storyflags:
 
 Sceneflags  
 
+Sealed Grounds:
+35: Sealed Grounds statue checked (in vanilla, this does nothing. Changed to toggle statue for starting statue rando)
```
lepelog, line 88: Imo this sceneflag was already used by the game for this purpose, just the statue select screen didn't check it, so I think this section can be removed

patches.yaml
```diff
@@ -29,7 +29,7 @@ global:
     - 115 # Full Lake Floria map
     - 129 # Updated Lanayru Desert map
     - 199 # talked to sword instructor after practice sword chest
-    - 804 # eldin entrance statue
+    # - 804 # eldin entrance statue - removed, as it is handled by starting statue rando
```
lepelog, line 32: I think the line can just be removed instead of commented out

patches.yaml
```diff
@@ -156,7 +156,7 @@ global:
       - 7  # pushed first minecart to end
       - 50 # Enter Lanayru Mining Facility (LMF) Cutscene
       - 55 # Lanayru Desert Intro Cutscene
-      - 68 # desert statue checked
+      # - 68 # desert statue checked - removed as it is handled by starting statue rando
```
lepelog, line 159: same here

logic/randomize.py
```diff
@@ -631,3 +632,58 @@ def randomize_dungeons_trials_starting_entrances(self):
...
+
+        # Starting bird statue rando
+
+        bsr = self.options["random-start-statues"]
+
+        possible_bird_statues = [
+            (entrance, values)
+            for entrance, values in self.areas.map_entrances.items()
+            if values.get("subtype", False)
+            and values["subtype"] == "bird-statue-entrance"
```
lepelog, line 644:
```suggestion
            if values.get("subtype") == "bird-statue-entrance"
```
comparing None is fine

lepelog (changes requested):

logic/randomize.py
```diff
@@ -631,3 +632,58 @@ def randomize_dungeons_trials_starting_entrances(self):
...
+            for entrance, values in possible_bird_statues
+            if values.get("province") == LANAYRU_PROVINCE
+        ]
+
+        lanayru_start_statue = self.rng.choice(possible_lanayru_statues)
+
+        # Logically bind the first-time dive to the statue to unlock it
+
+        self.placement.map_transitions[
+            "\Sky\\South East\\Faron Pillar\\First Time Dive"
```
yannl35133, line 676: Don't use these, they are not fixed names. This should be `self.areas.short_to_full(FARON_FIRST_TIME_DIVE)` with `FARON_FIRST_TIME_DIVE = Sky - Faron Pillar - First Time Dive` defined in `constants.py`

yannl35133 (approved):
lgtm

PR #526, opened 2024-01-14
Restructure Custom Functions

elijah-thomas774:


CovenEsme:
YES. Finally. I've been meaning to do this for ages. Thanks for taking the time to do this 🏆 

Looks good to me :p

asm/custom-functions/src/rando/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+#[link_section = "data"]
+static mut IS_FILE_START: bool = false;
+
+#[link_section = "data"]
+#[no_mangle]
+static mut FORCE_MOGMA_CAVE_DIVE: bool = false;
+
+#[no_mangle]
+pub fn process_startflags() {
+    (unsafe { *file_manager::get_ptr() }).anticommit_flag = 1;
```
lepelog, line 38:
```suggestion
    unsafe { (*file_manager::get_ptr()).anticommit_flag = 1 };
```
Otherwise rust just ignores this code, probably because it's UB somehow?

lepelog (changes requested):
Looks good, found just one instance where the newly generated was different

lepelog:
Yea, that's a good idea. However, I think that should be done in a separate PR, this is already a huge improvement that I don't want to just sit around for a while

PR #535, opened 2024-01-27
change split hero mode options to use a custom function instead of conditional patches

lepelog:
Just a small refactoring, doesn't change behavior


CovenEsme:
Tested with each option on/off and with a damage multiplier of 1x, 2x, and 5x. All worked as expected

PR #539, opened 2024-02-06
Make triforces and heart pieces whippable

CovenEsme:
Allows the whip to pickup Triforces and Heart Pieces. I've tested to make sure each patch is necessary to make each item whippable. It seems like there's a combination of defenderCollider flags that need to be set to allow items to be whippable.


YourAverageLink:
Lgtm

PR #546, opened 2024-02-15
Use Wch macro for some constant wide strings

robojumper:
Just to make things a bit more readable. wchz would also work for automatically zero-terminating the first two, but the last three contain interior NUL wide chars.


elijah-thomas774:
Looks good to me as long as it’s still byte equivalent (haven’t checked myself, but assuming  it is).  I could probably look in the future for more text commands to make it clearer in these  predefined strings.

PR #558, opened 2024-03-01
🔔 Academy Bell 🔔

CovenEsme:
## What does this PR do?
Randomizes the item dropped by the Knight Academy bell when it is rung. When the dropped item is collected by the player, Skyloft sceneflag 101 will be set. This flag is checked to make sure that the item cannot be collected multiple times.

* I tested that ringing the bell drops the randomized item
* I tested that you can ring the bell with a bomb throw (needs the trick enabled to be logically relevant)
* I tested that only one item spawns - even if you hit the bell multiple times
* I tested that hitting the bell, not collecting the item, reloading the area, and then hitting it again still spawns the randomized item
* I tested that hitting the bell, collecting the item, reloading the area, and then hitting it again *does not* spawn the randomized item
* I tested that you can still end the Archery minigame by shooting the academy bell (both if you have collected the item or if you haven't)

## Notes
Shooting the Knight Academy bell during the Archery minigame *cannot* be used to peak the bell item during the minigame. The camera pans away before the item falls out of the bell enough to see it :p


checks.yaml
```diff
@@ -25,6 +25,15 @@ Upper Skyloft - Sparring Hall Chest:
   type:
   Paths:
     - stage/F009r/r0/l0/TBox/66
+Upper Skyloft - Ring Knight Academy Bell:
+  original item: Green Rupee
+  type:
+  Paths:
+    - stage/F000/r0/l2/Bell/0xFDE8
```
lepelog, line 32: There is only one Bell on each layer, right? In those cases I have always omitted the id

lepelog (changes requested):

checks.yaml
CovenEsme: Yeah, there's just the one :p

CovenEsme: Fixed

lepelog (changes requested):

changedflags.md
```diff
@@ -82,10 +82,13 @@ New Trial Completed Storyflags:
 
 953: Group of Tadtones Counter
 
-Sceneflags  
+Sceneflags
 
-Ancient Cistern:  
-85: After receiving Item from flame CS  
+Skyloft:
+101: Collected item from Knight Academy Bell
```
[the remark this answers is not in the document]
CovenEsme: I made this change and tested that the new flag gets used instead of the old :p

lepelog (approved):

PR #559, opened 2024-03-03
Allow cleanly cancelling randomization

robojumper:
## What does this PR do?
Previously you could dismiss the progress dialog using ESC while randomization was ongoing, so you could have multiple randomization threads going on at the same time, usually resulting in a corrupt ISO.

This change allows randomization to be cleanly cancelled with a cancel button or the ESC key. This is achieved by throwing an exception in the progress callback after cancellation has been requested, which will fail the randomizer thread. The progress dialog is only ever manually closed by the main GUI thread after randomization has succeeded or failed.

* I tested on Windows
* I tested that the initial extraction progress dialog does **not** have a cancel button and cannot be cancelled via ESC.
* I tested that the randomization progress dialog has a cancel button.
* I tested that pressing this cancel button or hitting ESC makes randomization error out with a "Randomization was cancelled" error.
* I tested cancelling when the randomizer is generating hints, where it takes a few seconds for cancellation to go through. The progress dialog is not dismissed until the error popup appears.
* I tested cancelling randomization while files are being patched.

## Notes

Since hints generation does not call the progress callbacks, cancellation will take multiple seconds when hints are being generated. This could be improved by calling the progress callback when calculating the SotS items and generating hints, which would allow generation to be interrupted in this step.

### DIFF
--- a/gui/dialogs/progressbar/progressdialog.py
+++ b/gui/dialogs/progressbar/progressdialog.py
@@ -1,7 +1,8 @@
 import random
 from typing import Optional
-from PySide6.QtCore import Qt, QTimer
-from PySide6.QtWidgets import QProgressDialog, QLabel
+from PySide6.QtCore import QEvent, QObject, Qt, QTimer
+from PySide6.QtGui import QCloseEvent
+from PySide6.QtWidgets import QProgressDialog, QLabel, QPushButton
 
 tips = [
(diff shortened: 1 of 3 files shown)


CovenEsme:
Additionally tested that you can cancel a seed, regen it without closing the gui, and then start the seed without it being all messed up

PR #562, opened 2024-03-05
implement alloc and print for custom functions

lepelog:
## What does this PR do?
Adds convenience stuff for rust custom functions, like an allocator and a print macro for the debug console

Write some rust code and find out if the game crashes :)

Currently, none of the new features are used, so the actual asm didn't change


asm/custom-functions/src/system/ios.rs
```diff
@@ -0,0 +1,37 @@
+use core::{
+    alloc::Allocator,
+    ffi::{c_int, c_uint, c_void},
+    ptr::NonNull,
+};
+
+extern "C" {
+    pub fn iosAllocAligned(heap: *const c_void, size: c_uint, align: c_int) -> *mut u8;
+    pub fn iosFree(heap: *const c_void, ptr: *mut u8);
+    pub static IOS_HEAP: *const c_void;
```
elijah-thomas774, line 10: unless I am missing something, these symbols are not defined in `original_symbols.txt` yet. Probably not the most important as I dont think this will be used much if at all?

asm/custom-functions/src/system/heap.rs
```diff
@@ -9,9 +13,26 @@ pub struct List {
     pub offset: u16,
 }
 
+#[repr(C)]
+pub struct HeapVtable {
```
elijah-thomas774, line 17: For this and future vtables, Would it be better to keep the function names as seen in ghidra? I think having the impl to call all the vtable functions as snake case, but for easier cross referencing I would think having the actual vtables as the original names.

elijah-thomas774:
looks good to me :D. I dont think using the (`C_GAME`/`CsGame`) update  as the game update hook will cause issues. It seems it will always exist if the StageManager exist, which should always exist if a game is loaded. If you meant to hook it to main game update, it would be the Reloader (`GAME`/`ScGame`) update function (`0x801ba754`).

PR #565, opened 2024-03-14
Fix Floria and treasure selling logic

yannl35133:


logic/requirements/Faron.yaml
```diff
@@ -68,7 +68,7 @@ Faron Woods:
     Great Tree - Near Underwater Hole: Water Dragon's Scale
     Great Tree - Platforms: Clawshots
     Clawshot Target Branch: Clawshots
-    Lake Floria Dive: Open Door to Lake Floria | Early Lake Floria Tricks
+    Lake Floria Dive: Open Door to Lake Floria | (Early Lake Floria - Fence Hop Trick & Practice Sword)
```
robojumper, line 71: I think you still need Sword | Bomb Bag | Clawshots | Beetle | Bow to actually reach this door from main Faron since you either need to cut the trees on the path leading up to the door or activate the rope.

logic/requirements/Faron.yaml
```diff
@@ -97,10 +97,10 @@ Faron Woods:
       Unlock Statue: Nothing
 
   Ledge to Lake Floria:
-    entrance: Push Log
+    entrance: Push Log | (Early Lake Floria - Fence Hop Trick & Practice Sword) | Early Lake Floria - Swordless Rope Floria Trick
```
robojumper, line 100: The Swordless Rope Floria Trick is performed in Faron Woods - Entry, no? I don't think you need access to the main Faron area for that trick.

logic/requirements/Faron.yaml
yannl35133: Are you sure that this list is complete?

robojumper: Reasonably? Slingshot doesn't work, Skyward Strike doesn't seem to work (but a sword allows you to cut down trees anyway), a precise Bomb throw trick works (but you can blow up the trees anyway).

yannl35133: Is the vine too far for the whip?

robojumper: Whip reaches far enough to reliably hit the rope slightly above the vine, which doesn't destroy vine.

If you angle the Skyward Strike correctly you can actually get it with a Skyward Strike (logically redundant, just cut down trees), but the Whip doesn't seem to reach far enough to be able to avoid the rope.

robojumper (approved):
LGTM, this actually allows backwards Lake Floria if the relevant tricks are enabled.

PR #573, opened 2024-04-28
Add possibility for removed locations, better item placement restriction

yannl35133:
No external changes, unless there's a bug

### DIFF
--- a/logic/logic.py
+++ b/logic/logic.py
@@ -34,8 +34,8 @@ class PoolExit:
 
 @dataclass
 class Placement:
-    item_placement_limit: Dict[EXTENDED_ITEM_NAME, EXTENDED_ITEM_NAME] = field(
-        default_factory=lambda: defaultdict(lambda: EIN(str()))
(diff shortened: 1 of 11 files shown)


logic/logic.py
```diff
@@ -34,8 +34,8 @@ class PoolExit:
 
 @dataclass
 class Placement:
-    item_placement_limit: Dict[EXTENDED_ITEM_NAME, EXTENDED_ITEM_NAME] = field(
-        default_factory=lambda: defaultdict(lambda: EIN(str()))
+    item_placement_limit: Dict[EXTENDED_ITEM_NAME, Tuple[List[str], List[str]]] = field(
```
lepelog, line 37: Can you please add a comment here what `Tuple[List[str], List[str]]` means?

Looking through the rest of the code it looks like the first list is positive (and "" means everything for some reason) and the second list is negative, which is why `DEFAULT_PLACEMENT_LIMIT` means everything.

PR #579, opened 2024-05-15
Unbork test suite, add logic settings tests

robojumper:
## What does this PR do?

Developers who want to add additional rando logic features (with additional settings)
have no way to verify that settings have the desired impact on requirements other than
generating a bunch of seeds and inferring likely requirements, which is error-prone.
The rando has a test suite, but the logic tests are broken and not run in CI.

This disables the broken tests, adds some modestly fast logic settings tests,
and runs them in CI (disabling tests that need an ISO extract for CI only).

`python -mpytest`

## Notes

The test scaffolding doesn't allow specific constructs like specific starting tables,
starting entrances, or specific required dungeons. This would be a reasonable
future extension, perhaps with tests for plandomizer.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,6 +28,8 @@ jobs:
       - name: Build distribution
         run: python3 -mPyInstaller ssrando.spec
       - name: Run tests
+        run: python3 -mpytest -m "not iso"
+      - name: Run seed reproducibility check
         run: |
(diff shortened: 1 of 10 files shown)


CovenEsme:
Looks good. A really good step towards bringing back automated testing. This is what I get when I run it which looks as I'd expect it too based on the code changes
image

PR #582, opened 2024-06-06
Load arcs from disc at runtime

lepelog:
## What does this PR do?
This stops patching all arcs in for treasure chests and instead let the game load the arc from disc when necessary. There should be no difference to the current behavior, if there is, that's probably a bug.

Play the game and pay special attention to the chest openings, it is necessary to play a full seed. I'm already happy if it doesn't crash.

## Notes
While this PR doesn't have any benefit for the end user, it's a fundamental building block for something like showing the correct item model for progressive items (stage size is not an issue if the arcs are only loaded when needed) and multiworld.
I didn't bother cleaning up the commits, there are quite a few changes which actually don't belong here, but I'll clean that up later and it shouldn't cause problems until then.


lepelog:
The chest animation desyncs when the arc has to be loaded from disc. I have an idea how to fix that, but until then it doesn't need testing

PR #594, opened 2024-07-29
don't allow placing SoT in the trial

lepelog:
## What does this PR do?
Vanilla already checks for night to disallow placing SoT, extend it to also check for trial as well

Have SoT and try to place it in the trial, it shouldn't work. Placing it (at day) on normal skyloft works normally.

I already tested this change on dolphin.

## Notes
This uses an unaligned memory read, which I'm 99% sure is fine.


goldendevil1711:
Tested on console and it worked.

PR #599, opened 2024-08-25
Beatable Only logic mode

yannl35133:
Add Beatable Only and Beatable Then Banned as logic modes. Needs testing


options.yaml
```diff
@@ -384,12 +384,15 @@
   type: singlechoice
   bits: 3
   choices:
-    # - Glitchless
-    - BiTless
+    - Normal
+    - Beatable Only
+    - Beatable Then Banned
```
lepelog, line 389:
```suggestion
    - All Locations Reachable
    - Beatable Only
```

PR #601, opened 2024-09-01
update brres.py to support extracting External data

elijah-thomas774:
## What does this PR do?
- Adds External Folder support to brres.py
- Adds the `get_all_paths` for brres files
- slightly changes the name reading since the string pool gives the size of the string (no loop search for null term required)

brres modification *should* remain untouched. if texture replacement still works, this should be good to go.

## Notes
Only reason I wanted to update here is for any other program that may reference this file to do something.


lepelog (approved):
I tested randomizing link's color and that worked

PR #605, opened 2024-10-24
Fix Chest on Top of LMF conditional hint behavior

goldendevil1711:
I think this is right but someone check the syntax to make sure I fixed it. The point is that the condition is that LMF is not Open *AND* LMF is Not Required.

## What does this PR do?
The fundamental code changes and what they translate to functionally.

How to test/recreate the issue/verify the changes.

## Notes
Any notes, additional thoughts, concerns or comments.


goldendevil1711:
Ran a few Dry Run test seeds. This location was an Always on Allsanity (DER on), EUD disabled, and without LMF (on modified Remlits settings). It was not hinted on modified Remlits when LMF was disabled

lepelog:
Replaced by #607

PR #608, opened 2024-11-14
Fix Flag Collision w/ Eldin Trial Relic 1

YourAverageLink:
## What does this PR do?
Since Relic scene flags were delegated from 102-111, Relic 1 in Eldin Trial would disappear after the first load, as its flag (102) is shared by the flag set at the start of the trial. This offsets all flags by 1 to fix this issue.

I loaded into each silent realm, verifying that none of the scene flags set collided with a relic's scene flag, and verified that no items disappeared after a reload.

## Notes
Also sets the intro text flags as startflags for each silent realm, skipping the Spirit Vessel animation & question prompt

### DIFF
--- a/checks.yaml
+++ b/checks.yaml
@@ -1858,52 +1858,52 @@ Skyloft Silent Realm - Relic 1:
   type: skyloft, silent realm
   Paths:
     # fake object name to handle them better
-    - stage/S000/r0/l2/Relic/102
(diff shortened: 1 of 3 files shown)


YourAverageLink:
This has been tested extensively in the Pre-S3 build, and no issues have been found.

PR #610, opened 2024-11-15
Comprehensive Song Hints

YourAverageLink:
## What does this PR do?
Adds a Comprehensive option to trial gate song hints. This works similarly to how song hints work in the HD Randomizer.
Comprehensive song hints have a "headline" item, that is, an item that is named in the hint.
- If there are any required (SotS) items in the silent realm, a random SotS item is chosen as the headline item
- If there are any possibly required items (see hint importance), a random possibly required item is chosen as the headline item
- Otherwise, a random item is technically chosen, but it is not shown in the hint (which simply states that the trial contains no useful items).
The hint also lists how many other useful (required or possibly required) items exist in the silent realm.

**This pull request depends on #597**

- Verify that all older song hint options still work as intended
- Verify that the comprehensive song hints give accurate information and prioritize SotS items correctly
- Verify that, with comprehensive song hints, silent realms cannot be hinted by any other hint, including SotS, Path, or Barren hints (as all of their locations count as being hinted already).

## Notes
For a racing setting, it may be desired to prioritize certain possibly required items (namely swords if the Gate of Time requires a sword lower than TMS), but such a feature will either be specific to a tournament version or become part of the hint distribution somehow.


YourAverageLink:
I'll mark this as a draft until #597 is merged

PR #611, opened 2024-11-17
Improve element layout for better counter legibility

CelestialKitsune:
## What does this PR do?
This moves all the QSpinBox elements below their respective QLabel so they can take the whole width of their column to properly display their value.

Only by launching the rando app, given this is only UI changes. 

## Notes
This has already been tested by someone who had issues before.


goldendevil1711:
I tested this, can confirm this works.

lepelog (approved):

PR #621, opened 2025-07-11
Add Option to shuffle Key Pieces in Eldin Volcano Only

YourAverageLink:
## What does this PR do?
Similar to https://github.com/mint-choc-chip-skyblade/sshd-rando/pull/532, this pull request amends the Open ET option to include a setting that will place Key Pieces randomly in Eldin Volcano, as though they were Own Dungeon small keys.

I've looked at spoiler logs with seeds with the Eldin Volcano shuffle option on and verified that the key pieces only show up there. More testing is needed for actual game patching (that the previous two options still work, for instance), and verification that Key Pieces can't be path hinted with Eldin Volcano shuffle since they're restricted items. I'll open this as a draft PR until those get more testing.

## Notes
Unlike the SSHDR PR, I have opted *not* to include Mogma Turf as a possible region for Eldin key piece shuffle. Discussion is welcome.


yannl35133 (approved):
Not sure why the gamepatches changed, but looks good to me

YourAverageLink:
> Not sure why the gamepatches changed, but looks good to me 

Must have been an update to the formatted because that was changed by running `black .`

PR #623, opened 2025-09-12
Add Option for Skip Harp Playing

gymnast86:
## What does this PR do?
This adds the option `Skip Harp Playing`. When this option is selected, trial gates will automatically open up when both the harp and song associated with the trial gate have been collected. 

I tested collecting each song both on the stage the associated trial gate appears and on different stages. The trial gates each opened as expected. I also tested starting with the Harp and all songs and all trial gates were open as expected.

### DIFF
--- a/asm/original_symbols.txt
+++ b/asm/original_symbols.txt
@@ -117,7 +117,6 @@ main.dol:
     getGratitudeCrystalCount: 0x80252710
     processSpecialItemDropIndex: 0x800c7bb0
     spawnDrop: 0x80248c90
-    setItemFlag: 0x80251e40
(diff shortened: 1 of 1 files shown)


gymnast86:
Is there somewhere I can find documentation on how the REL header is laid out?

gymnast86:
I decided to instead implement the checks in asm at the end of the `dAcOWarp::executeState_GateWait` function which runs every frame that a trial gate is loaded, but hasn't been opened yet. I also realized that the flag for the trial gate has to be set globally for it to appear visually correct when the harp/song is obtained on the same stage as the trial gate when it's already loaded.

